PR #56, opened 2021-05-15
refact(env): Change default registry for linux-utils image to default registry (docker.io)

niladrih:

Uses docker as the default registry for the linux-utils image used in localpv helper pods. This is to keep with the convention of using docker registry by default.

Change k8s version prerequisite to 1.16 based on the availability of CustomResourceDefinition on apiextension.k8s.io/v1 (BlockDevice and BlockDeviceClaim CRs from github.com/openebs/node-disk-manager repo).
Ref: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

**What this PR does?**:
Changes the string value of the default linux-utils image variable.
Docs change.

No.

### DIFF
--- a/cmd/provisioner-localpv/app/env.go
+++ b/cmd/provisioner-localpv/app/env.go
@@ -45,7 +45,7 @@ const (
 )
 
 var (
-	defaultHelperImage = "quay.io/openebs/linux-utils:latest"
+	defaultHelperImage = "openebs/linux-utils:latest"
 	defaultBasePath    = "/var/openebs/local"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/provisioner-localpv/app/env.go
```diff
@@ -45,7 +45,7 @@ const (
 )
 
 var (
-	defaultHelperImage = "quay.io/openebs/linux-utils:latest"
+	defaultHelperImage = "openebs/linux-utils:latest"
```
kmova, line 48: 👍

kmova (approved):

akhilerm (approved):
LGTM

PR #64, opened 2021-06-28
chore(tests): Add BDD tests

niladrih:
Signed-off-by: Niladri Halder <niladri.halder@mayadata.io>

This PR adds BDD tests based on GInkgo + Gomega for localpv-provisioner.

PR #67, opened 2021-07-05
refact(integration-test): Rework integration tests and ci operations on integration tests

niladrih:

Change summary:
1. Add build files for StorageClass to pkg/ directory
2. Removes dependency on github.com/openebs/maya/tests/artifacts package
3. Refactors delete PV/PVC logic
4. Delete logic now checks for multiple parameters before moving on to the next test spec.
5. Refactors after-test cleanup logic
6. Implements sort-by-timestamp for BDC (commented -- because it does not have any use, is left because it can turn out to be useful)
7. Add makefile targets for separate hostpath and device integration testing.
8. All integration tests (hostpath + device) run in GitHub Actions workflows on PR raise and master build.


kmova:
This PR needs to be updated with changes to make use of the new builder files in the BDD. Depends on #64

kmova (approved):

kmova:
Good work @niladrih !!

Issue #13, opened 2020-11-23
HostPath does not enforce quotas


jshin47:
**OpenEBS LocalPV HostPath do not enforce quotas**
This is a known limitation of OpenEBS LocalPV HostPath, but it's one that seriously limits its usefulness in production.

**I'd like for OpenEBS LocalPV HostPath to collect disk usage stats and render pods unscheduleable when they exceed quotas**
I know this is not as straight forward as it sounds, but I'd like to make OpenEBS HostPath render pods unschedulable when quotas are exceeded. For existing pods, I'd like to kill them if they are exceeding quotas. For pods to be scheduled, I'd like for this quota to be enforced in advance.

**Anything else you would like to add:**
I know this is a roadmap feature.

kmova:
Started experimenting with the feature of creating a sparse file within the hostpath to enforce quota. (inline with what Rawfile Local Pv does). 

The PV spec will be changed to:

```
    local:
      fsType: ""
      path: /var/openebs/local/pvc-4ae3d213-0a93-4b77-9851-d5af29fc07f1/disk.img
```

The above results in Kubelet rejecting to mount the device with the following error: 
```
MountVolume.NewMounter initialization failed for volume "pvc-4ae3d213-0a93-4b77-9851-d5af29fc07f1" : only directory and block device are supported
```

Need to figure out if I can easily convert the sparse file into a block device. 

The current code is located at: https://github.com/kmova/dynamic-localpv-provisioner/tree/enforce-quota

jshin47:
Thanks! I'll see if I can help in any way. I doubt it, but I'd like to learn how to set up a dev environment for this anyway because we may be using it heavily.

One thing that comes to mind is to use `losetup -fP` and then use that loopback device as your block device

kmova:
Another related question with hostpath, which filesystem do you plan to use - xfs, ext4 or anything else. 

Specifically, would it be ok to use XFS - as there is a possibility of using set quota on top of XFS.

ankur-gupta-guavus:
I prefer using XFS only

jshin47:
I use XFS only as well.

kmova:
This feature also needs to account for
- volume resize 

This feature needs a good implementation of generating unique xfs project ids per node, that are also unique across clusters in cases the PVs are migrated/merged onto same node.

almas33:
Hi, I have submitted an application and excited to work on this. 😄

kmova:
[...]

PR #78, opened 2021-08-08
feat(hostpath):enforce xfs quota on the hostpath

almas33:
## feat(hostpath):enforce xfs quota on the hostpath

Issue no. - #13 

To test:-

1. Follow  documentation of task 1 - https://docs.google.com/document/d/1Vs-8Xv4fm_B5-qp4I7dYEVYfO4hEVY7KMV9uhAgdZfk/edit# to setup xfs fs in the basepath.
2. Use my image `almas33/openebs-pv` in place of `openebs/provisioner-localpv` and `almas33/xfs-quota` in place of `openebs/linux-utils`
3. Add `type:enforceXfsQuota` , `bsoft` as soft limit and `bhard` as hard limit in the parameters in storage class.yaml
4. Run pod as done normally.


cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -39,6 +39,7 @@ import (
 type podConfig struct {
 	pOpts                         *HelperPodOptions
 	parentDir, volumeDir, podName string
+	bsoft, bhard                  string
```
Ab-hishek: Instead of making this a part of `podConfig` we should make it a part of `helperPodOptions`struct.

cmd/provisioner-localpv/app/provisioner_hostpath.go
```diff
@@ -87,6 +87,19 @@ func (p *Provisioner) ProvisionHostPath(ctx context.Context, opts pvController.P
 		return nil, pvController.ProvisioningFinished, iErr
 	}
 
+	bsoft := opts.StorageClass.Parameters["bsoft"]
+	bhard := opts.StorageClass.Parameters["bhard"]
```
Ab-hishek: Declare constants for bsoft, bhard.

cmd/provisioner-localpv/app/provisioner_hostpath.go
```diff
@@ -87,6 +87,19 @@ func (p *Provisioner) ProvisionHostPath(ctx context.Context, opts pvController.P
...
+
+	if bsoft != "" ||
+		bhard != "" {
```
Ab-hishek: Is the code formatted??

cmd/provisioner-localpv/app/helper_hostpath.go
Ab-hishek: We also add a parameter called `type: enforceXfsQuopta` in the storageclass parameter to let users know that this storageclass is for enforcing quotas for xfs fs.

cmd/provisioner-localpv/app/provisioner_hostpath.go
Ab-hishek: Better way to check for empty string would be `len(bsoft) != 0 || len(bhard) != 0`.

Ab-hishek: We will also need to validate the limits before creating the pods. The possible validations would include:
1. bsoft should not be greater than bhard.
2. bsoft and bhard cannot be in decimals.
3. If someone provides the values as MI or Gi then we need to change it from Mi to M as limits would work only for M, G, m, g types.

cmd/provisioner-localpv/app/provisioner_hostpath.go
Ab-hishek: And add them to the `podOpts` struct instead of passing them as parameters inside the function.

cmd/provisioner-localpv/app/quota_hostpath.go
```diff
@@ -0,0 +1,151 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+This code was taken from https://github.com/rancher/local-path-provisioner
+and modified to work with the configuration options used by OpenEBS
+*/
+
+package app
+
+import (
```
Ab-hishek: IMO I don't think a new file is necessary. We can have it inside the helper-hostpath.go file inside itself with some refactoring as the create and delete pod functions are the same, we can use the existing ones itself.

cmd/provisioner-localpv/app/quota_hostpath.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (p *Provisioner) launchQuotaPod(ctx context.Context, config podConfig) (*corev1.Pod, error) {
+	// the quota pod need to be launched in privileged mode. This is because in CoreOS
+	// nodes, pods without privileged access cannot write to the host directory.
+	// Quota pods need to setup projectID which requires to write the mapping of projectID and subdirectory.
+
+	lastPid := "PID=`xfs_quota -x -c 'report -h' /data | tail -2 | awk 'NR==1{print substr ($1,2)}+0'` ;"
+	newPid := "PID=`expr $PID + 1` ;"
+	initializeProject := "xfs_quota -x -c 'project -s -p " + filepath.Join("/data/", config.volumeDir) + " '$PID /data ;"
+	setQuota := "xfs_quota -x -c 'limit -p bsoft=" + config.bsoft + " bhard=" + config.bhard + " '$PID /data"
```
Ab-hishek: We also will need to account for where the users only provides any one the limits- either `bosft` or `bhard`. In that case the command will fail but the limits are valid as we can also have either one of them set. It is not necessary to provide both.

Ab-hishek: I am not sure whether this will work on only system as while testing myself I was not able to run these commands without the root access. So, I feel better to give this pod root access. cc: @kmova @akhilerm .

cmd/provisioner-localpv/app/quota_hostpath.go
```diff
@@ -0,0 +1,151 @@
...
+
+	config.pOpts.cmdsForPath = []string{"sh", "-c", lastPid + newPid + initializeProject + setQuota}
+
+	privileged := true
+
+	klog.Infof("Setting the project quota...")
+	klog.Infof("Check filesystem of mount if quota didn't set")
```
Ab-hishek: Where are we checking the fs of the volume?

cmd/provisioner-localpv/app/quota_hostpath.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	var qPod *corev1.Pod
+
+	//Launch the quota pod.
+	qPod, err = p.kubeClient.CoreV1().Pods(p.namespace).Create(ctx, quotaPod, metav1.CreateOptions{})
+	return qPod, err
+}
+
+func (p *Provisioner) exitQuotaPod(ctx context.Context, qPod *corev1.Pod) error {
```
Ab-hishek: We can use the same function present in the helper_hostpath.go file instead making a redundant copy of this function.

cmd/provisioner-localpv/app/quota_hostpath.go
Ab-hishek: Also in this way, if one commands fails then what's the point of running the subsequent commands. I feel rather than passing these commands like this better to create a shell script for all these commands(with proper error or successful logs) but still the logs in the pod won't be of much help as the pod will be deleted after few seconds and pass them as configmap volumes to the helper pod and execute the script. cc: @kmova @akhilerm .

Ab-hishek (changes requested):
@almas33 Given some comments for changes.

cmd/provisioner-localpv/app/provisioner_hostpath.go
```diff
@@ -87,6 +87,19 @@ func (p *Provisioner) ProvisionHostPath(ctx context.Context, opts pvController.P
...
+
+		qErr := p.createInitQuotaPod(ctx, podOpts, bsoft, bhard)
+		if qErr != nil {
+			klog.Infof("Setting quota failed: %v", name, qErr)
+			return nil, pvController.ProvisioningFinished, qErr
+		}
+	}
```
Ab-hishek: Add logs to indicate the success scenario also.

cmd/provisioner-localpv/app/quota_hostpath.go
almas33: Can you confirm this we can use that commands in helper pod?
I also think it is unnecessary to launch a pod.

Ab-hishek: No, I meant creating a new pod is necessary in cases where the user's want quota. If the user doesn't want to enforce quota then we will not launch the quota pod. What I meant is shift the code of this file in the helper_hostpath.go file itself and instead of creating a new function for deleting the pod you can call the same function and for creating also same function can be called with some additional parameters added if needed.

cmd/provisioner-localpv/app/helper_hostpath.go
almas33: Done

cmd/provisioner-localpv/app/provisioner_hostpath.go
almas33: Done

cmd/provisioner-localpv/app/provisioner_hostpath.go
almas33: most of the possible validations are now covered

almas33: we can also use decimal values now

cmd/provisioner-localpv/app/quota_hostpath.go
almas33: covered in `validateLimits()`

cmd/provisioner-localpv/app/quota_hostpath.go
almas33: Done

cmd/provisioner-localpv/app/provisioner_hostpath.go
almas33: Added logs

cmd/provisioner-localpv/app/quota_hostpath.go
almas33: Its not possibe to check fs without launching pod 
can we do something similar to the way we are checking :- if volumeDir is not under root?

cmd/provisioner-localpv/app/provisioner_hostpath.go
almas33: > Is the code formatted??

Is this not correct format?
In my system there is small gap but after pushing it becomes big , it may be related to `Tab` spaces or it auto formats beacuse it can only be seen in `if` statements

Ab-hishek: > we can also use decimal values now

No, we should not use decimals as there is a limitation for decimal values while applying quotas.` The limits cannot be in decimals`.  That's what the check needs to be- check whether the limits are not given in decimals because that would cause errors while enforcing quotas.

Ab-hishek: > Is this not correct format?
In my system there is small gap but after pushing it becomes big , it may be related to Tab spaces or it auto formats beacuse it can only be seen in if statements

Just run `make fmt` command from the root of the project folder.

almas33: > > we can also use decimal values now
> 
> No, we should not use decimals as there is a limitation for decimal values while applying quotas.` The limits cannot be in decimals`. That's what the check needs to be- check whether the limits are not given in decimals because that would cause errors while enforcing quotas.

Yes , I am fully aware that we can't use decimal values (in the command) that's why I added a functionality so that we can use decimal values (in the parameters) . You can see `convertToK` function which I added in recent commit

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -92,6 +104,63 @@ func (pOpts *HelperPodOptions) validate() error {
 	return nil
 }
 
+//validateLimits check that the limits to setup qouta are valid
+func (pOpts *HelperPodOptions) validateLimits() error {
+	if pOpts.bsoft == "0k" &&
+		pOpts.bhard == "0k" {
+		return errors.Errorf("both limits cannot be 0")
```
Ab-hishek, line 111: Return errors in detailed format
```suggestion

klog.Infof("Setting xfs quota for volume %v failed: %v", name, err)
				alertlog.Logger.Errorw("",
					"eventcode", "local.pv.quota.failure",
					"msg", "Failed to set xfs quota limits for Local PV",
					"rname", opts.PVName,
					"reason", "xfs quota validation failed",
					"storagetype", stgType,
				)
				return nil, err
```

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -92,6 +104,63 @@ func (pOpts *HelperPodOptions) validate() error {
...
+
+	if pOpts.bsoft == "0k" ||
+		pOpts.bhard == "0k" {
+		return nil
+	}
+
+	if len(pOpts.bsoft) > len(pOpts.bhard) ||
+		(len(pOpts.bsoft) == len(pOpts.bhard) &&
+			pOpts.bsoft > pOpts.bhard) {
+		return errors.Errorf("hard limit cannot be smaller than soft limit")
```
Ab-hishek, line 122: Similar suggestion here and for every other places where errors are handled.

cmd/provisioner-localpv/app/provisioner_hostpath.go
```diff
@@ -87,6 +93,43 @@ func (p *Provisioner) ProvisionHostPath(ctx context.Context, opts pvController.P
...
+			name:                   name,
+			path:                   path,
+			nodeAffinityLabelKey:   nodeAffinityKey,
+			nodeAffinityLabelValue: nodeAffinityValue,
+			serviceAccountName:     saName,
+			selectedNodeTaints:     taints,
+			imagePullSecrets:       imagePullSecrets,
+			bsoft:                  bsoft,
+			bhard:                  bhard,
+		}
```
Ab-hishek: All these can be performed outside of this if condition. If the fields are not filled by the user then they will remain empty. And for if condition you can check for:
```suggestion
		if scType == "enforceXfsQuota" && (len(Bsoft) != 0 || len(Bhard) != 0)
```
And then have proper error handling checks too.

cmd/provisioner-localpv/app/quota_hostpath.go
Ab-hishek: No, there is no way we can do this from the provisioner. This will have to be from the helper pod itself.

cmd/provisioner-localpv/app/provisioner_hostpath.go
almas33: I don't think we should check like this `if scType == "enforceXfsQuota" && (len(Bsoft) != 0 || len(Bhard) != 0)
` because if user has given type `enforceXfsQuota` that means he wants to set quota , then we should try to create quota pod , but while pod creation the error `both limits can't be 0 ` will be displayed and it will fail but doing like the way you suggested there will be no error displayed

almas33: If we want to remove redeclaration of `podOpts` we can declare value of `bsoft` and `bhard` while sending it to helper pod , it will be empty anyways ( values are not provided in parameters )as variable is declared in `podOpts`.

cmd/provisioner-localpv/app/helper_hostpath.go
almas33: `PVName` is not available in `helper_hostpath.go` so I should skip it , right?

almas33: and `stgType` is also not available

Ab-hishek: Okay. Then leave it. Since in provisioner_hostpath.go the logs are properly errored out.

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -92,6 +104,63 @@ func (pOpts *HelperPodOptions) validate() error {
...
+	}
+
+	return nil
+}
+
+//converToK converts the limits to kilobytes
+func convertToK(limit string) (string, error) {
```
kmova: @almas33 - can you add unit tests for this function?

cmd/provisioner-localpv/app/provisioner_hostpath.go
kmova: @almas33 - it is quite possible to misconfigure the parameters or provide incorrect parameters by users - while they are learning how to use the solution. The code should handle this via:
* Sane defaults
* Catch errors

almas33: @kmova - does that mean there will be some default limits applied if there is error in limits entered in parameter . Error handling of incorrect parameters are already taken care in `convertToK()`

kmova: @almas33, let us get back to this comment after the changes suggested in the design doc for storage class parameters are updated. The validation will change.

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -50,8 +50,9 @@ var (
 	//to be launched.
 	CmdTimeoutCounts = 120
 
-	//conversion stores the equivalent of 1kb stored as 1kb, 1mb, 1gb, 1tb
-	conversion = []float64{1, 1024, 1048576, 1073741824}
+	//conversion stores the size as 1kb, 1mb, 1gb, 1tb
+	//respectively in kb
+	conversion = []float64{math.Pow(2, 0), math.Pow(2, 10), math.Pow(2, 20), math.Pow(2, 30)}
```
Ab-hishek: This is not used anywhere? Is this array required?

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -106,59 +110,56 @@ func (pOpts *HelperPodOptions) validate() error {
...
 
 	return nil
 }
 
 //converToK converts the limits to kilobytes
-func convertToK(limit string) (string, error) {
+func convertToK(limit string, pvcStorage int64) (string, error) {
 
 	if len(limit) == 0 {
 		return "0k", nil
```
Ab-hishek, line 132: Instead of "OK", you can return null(which would signify empty value)

cmd/provisioner-localpv/app/helper_hostpath.go
almas33: No, I will remove it . It was used previously

cmd/provisioner-localpv/app/helper_hostpath.go
almas33: its not "OK" it is "0k" , it is required because bsoft or bhard doesn't accept null

Ab-hishek: So instead of `0K` return empty string i.e "". In that case we'll have better picture that the values were empty and not provided as 0.

cmd/provisioner-localpv/app/helper_test.go
```diff
@@ -0,0 +1,59 @@
...
+			value:       "200%",
+			expectValue: "10000000k",
+		},
+	}
+
+	for k, v := range testCases {
+		v := v
+		t.Run(k, func(t *testing.T) {
+			if len(v.value) != 0 {
+				os.Setenv(string(menv.OpenEBSNamespace), v.value)
```
Ab-hishek: Why do we need to set the env variable for this function?

cmd/provisioner-localpv/app/helper_test.go
```diff
@@ -0,0 +1,59 @@
...
+			}
+			actualValue, _ := convertToK(v.value, 5000000000)
```
Ab-hishek: Also I see you have hardcoded the pvcStorage variable value. Better to follow the template that go provides. You can add testcases in the similar fashion. What needs to done is if you are using goland IDE:
1. right click on the function to be tested
2. click on generate
3. click on test for this function
A template for the test function will be automatically generated then fill the values for test cases and any changes that is required.

cmd/provisioner-localpv/app/helper_hostpath.go
almas33: the command will give error it only accepts size , if you do `bsoft=0k` that means soft limit will not be applied, there is no limit as `0k`. Providing value as 0 automatically means no limit applied

Ab-hishek: I don't think that will reach that stage. Since if both the limits are empty then the validateLimits() will return error and we will not reach the stage of executing the quota pod. Here you are not checking the `0` value but checking the length of the limit string which if 0 means the user did not provide any value for that limit.

Ab-hishek: Either you can address this or else I am fine with this also.

almas33: If we address this code will unnecessarily become complex and for consistency , I think this is fine

cmd/provisioner-localpv/app/helper_test.go
almas33: We don't

Ab-hishek:
The PR looks good to me. @kmova can you review it once.

Ab-hishek:
I'll try to add integration tests in the same PR. Or else, we can take that up in a separate PR.

Ab-hishek (approved):

niladrih:
@almas33 Could you please rebase your branch with 'develop' once? Thanks.

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -165,6 +240,72 @@ func (p *Provisioner) createCleanupPod(ctx context.Context, pOpts *HelperPodOpti
...
+		ExtractSubPath()
+	if vErr != nil {
+		return vErr
+	}
+
+	//Pass on the taints, to create tolerations.
+	config.taints = pOpts.selectedNodeTaints
+
+	var lErr error
+	config.pOpts.softLimitGrace, lErr = convertToK(config.pOpts.softLimitGrace, config.pOpts.pvcStorage)
```
kmova, line 269: When the values are not specified. convertToK will return "0k" and that will be passed down to the quota command below. If limit is set at `0k` - does it become unlimited? or how is the enforcement made?

almas33: @kmova  - If anyone of the limit is set to 0k then enforcement will be made using only one of the limit
if both limits are 0k then quota enforcement is not possible . That case is covered in `validateLimits()`

PR #91, opened 2021-08-27
docs(enable-xfs-quota) - documentation for applying xfs quota

almas33:
Added .md file of a quickstart documentation for setting xfs quota on the Hostpath

Signed-off-by : Almas Ahmad <ahmadalmas.786.aa@gmail.com>

### DIFF
--- a/docs/enable-xfs.md
+++ b/docs/enable-xfs.md
@@ -1,0 +1,146 @@
+# OpenEBS LocalPV Hostpath Enable Quota
+
+### Prerequisites
+
+1. A Kubernetes cluster with Kubernetes v1.16 or above is required
+2. All the nodes must have xfs utils installed
+3. The base path used by the provisioner should be mounted with XFS project quotas enabled
+
+### Setup for testing
+Create a disk file which will be formatted as xfs fs, mounted as loopback device and exposed as the Basepath
+```console
+sudo dd if=/dev/zero of=xfs.32M bs=1 count=0 seek=32M
+sudo mkfs -t xfs -q xfs.32M
+sudo mount -o loop,rw xfs.32M -o pquota /var/openebs/local
+```
+
+### Check Basepath
+Verify that filesystem is xfs
+```console
+stat -f -c %T /var/openebs/local
+```
+Filesystem of Basepath will be shown 
+```console
+xfs
+```
+
+Verify that project quotas are enabled
+```console
+xfs_quota -x -c state | grep 'Project quota state on /var/openebs/local' -A 2
+```
+Both Accounting and Enforcement should be ON
+```console
+Project quota state on /var/openebs/local (/dev/loop16)
+  Accounting: ON
+  Enforcement: ON
+```
+
+### Installing
+Install the OpenEBS Dynamic LocalPV Provisioner using the following command:
+```console
+kubectl apply -f https://openebs.github.io/charts/openebs-operator-lite.yaml
+```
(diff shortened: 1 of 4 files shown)


docs/enable-xfs.md
```diff
@@ -0,0 +1,146 @@
...
+reclaimPolicy: Delete
+parameters:
+  enableXfsQuota: 'true'
+  softLimitGrace: 20%
+  hardLimitGrace: 40%
+```
+`softLimitGrace` and `hardLimitGrace` with PV Storage Request will decide the soft limit and hard limit to be set.
+The size of a limit will be => size of PV Storage request ( 1 + LimitGrace% )
+Anyone of hard limit or soft limit can also be used
+[Click here](https://man7.org/linux/man-pages/man8/xfs_quota.8.html) and see Quota Overview section for detailed instructions about soft and hard limits.
```
Ab-hishek: You can directly attach the link to the quota overview section, instead of mentioning it separating.
```suggestion
[Click here](https://man7.org/linux/man-pages/man8/xfs_quota.8.html#QUOTA_OVERVIEW)  for detailed instructions about soft and hard limits.
```

Ab-hishek (approved):
One minor comment. Otherwise PR looks good.

niladrih:
1. Could you please move to a directory as described below...
/docs/tutorials/hostpath/<doc>

2. Could you rename this doc as...
        a. 'enable-xfs-quota.md' -- if the content available at the moment is the extent of this doc, and you don't plan to add instructions to disable the quota options. If you want, then along with this, you may create a doc with a name 'disable-xfs-quota.md' to include the disable instructions.
        b. 'xfs-quota.md' -- if you plan to add disable instructions as well

niladrih:
@almas33 Because this is meant to be for setups with native XFS filesystems, could you remove the part where the loop device is created and mounted?

Could you change the title of the document to 'OpenEBS LocalPV Hostpath Enable XFS Quota'

Could you add 'XFS filesystem' as a prerequisite for this?

Could you add commands for a user to check the mount options of the base filesystem?

Could you also add a command to remount the existing filesystem with 'pquota' option included in case the user finds that his mount does not have that option?

Also please remove the comments in the sample StorageClass, as it is looks a bit crowded.

cc: @Ab-hishek

Ab-hishek:
> could you remove the part where the loop device is created and mounted?

IMO this can be kept as a separate setup step for those users who want to test this feature but doesn't have native xfs fs. Similar to what we have currently for lvm-localpv. cc: @niladrih

niladrih:
@Ab-hishek -- Yup. That works. The main go-to instruction set should not have the loop device instructions, IMO.
If you plan to include 2 files, put them under a directory like so...
docs/tutorials/hostpath/xfs_quota/

You can name the 2nd file appropriately --> be sure the include the keywork 'loop' in your name.

almas33:
> Could you add commands for a user to check the mount options of the base filesystem?

I have already added this , not sure what you meant by this.

almas33:
@niladrih , I have added instructions to enable pquota if xfs fs doesn't have , but I don't have xfs fs , so I was unable to check
let me see if I can check somehow that it is working or not

niladrih:
@almas33 -- I can verify the instructions for you. I'll get back to you in this thread.

niladrih (approved):
@almas33 -- Changes look good Ubuntu/Debian temporary usage. LGTM.

PR #96, opened 2021-09-03
test(xfs-hostpath-quota): Add integration tests for xfs hostpath quota enforcement

Ab-hishek:
Signed-off-by: Abhishek Agarwal <abhishek.agarwal@mayadata.io>

This PR adds BDD tests for xfs-quota feature.

**What this PR does?**:
The BDD tests for this feature needs:
1.  `xfsprogs` dependency to be present in the system beforehand.
2.  `sudo` access for some infrastructure setup(loopback device creation, dir inside root folder creation, mounting the loopback device) to run these tests

No.

The BDD tests covers 2 scenarios:
1. When the underlying fs is not `xfs` and quota storageclass is used for provisioning.
2. When the underlying fs is `xfs` and a valid quota storageclass is used for provisioning.

**Checklist:**
- [ ] Fixes #<issue number>
- [x] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests
- [x] Commit has integration tests


Makefile
```diff
@@ -142,7 +142,7 @@ testv: format
 # Requires KUBECONFIG env
 .PHONY: integration-test
 integration-test:
-	@cd tests && go test -ginkgo.v
+	@cd tests && go test -ginkgo.v -timeout 1200s
```
mynktl: nit: better to convert 1200s -> 20m

pkg/kubernetes/api/core/v1/pod/build.go
```diff
@@ -90,6 +91,27 @@ func (b *Builder) WithNamespace(namespace string) *Builder {
...
+	}
+
+	// copy of original map
+	newmatchlabels := map[string]string{}
+	for key, value := range matchlabels {
+		newmatchlabels[key] = value
+	}
+
+	// override
+	b.pod.object.Labels = newmatchlabels
```
mynktl: here we will lose existing labels

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
+/*
+Copyright 2019 The OpenEBS Authors
```
mynktl:
```suggestion
Copyright 2021 The OpenEBS Authors
```

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+		imageName: generateDiskImageName(),
+		Name:      "",
+	}
+	return disk
+}
+
+// Generates a random image name for the backing file.
+// the file name will be of the format fakeXXX, where X=[0-9]
+func generateDiskImageName() string {
+	rand.Seed(time.Now().UTC().UnixNano())
```
mynktl, line 66: nit: we don't need to seed frequently. we can move this function to main/init function

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+// testing.
+type Disk struct {
+	// Size in bytes
+	Size int64
+	// The backing image name
+	// eg: /tmp/fake123
+	imageName string
+	// the disk name
+	// eg: /dev/loop9002
+	Name string
```
mynktl: nit: can we rename it to `DiskPath` for better readability?

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+func generateDiskImageName() string {
+	rand.Seed(time.Now().UTC().UnixNano())
+	randomNumber := 100 + rand.Intn(899)
+	imageName := "fake" + strconv.Itoa(randomNumber)
+	return imageDirectory + "/" + imageName
+}
+
+// AttachDisk triggers a udev add event for the disk. If the disk is not present, the loop
+// device is created and event is triggered
+func (disk *Disk) AttachDisk() error {
```
mynktl: comment seems confusing. This function is not triggering udev event for disk, It is creating only loop device for the given path. If this fn is specific to creating loop device for a file then we can rename this function to CreateLoopDevice or something else.

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (disk *Disk) Unmount() error {
+	var lastErr error = nil
+	for _, mp := range disk.MountPoints {
+		err := RunCommandWithSudo("umount " + mp)
+		if err != nil {
+			lastErr = err
+		}
+	}
+	return lastErr
```
mynktl, line 177: lets remove mount entry from `disk.MountPoints` also

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// DetachAndDeleteDisk detaches the loop device from the backing
+// image. Also deletes the backing image and block device file in /dev
+func (disk *Disk) DetachAndDeleteDisk() error {
+	if disk.Name == "" {
```
mynktl: better to add check if disk is mounted or not

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
+/*
+Copyright 2019 The OpenEBS Authors
```
mynktl:
```suggestion
Copyright 2021 The OpenEBS Authors
```

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+				Expect(err).To(
+					BeNil(),
+					"while creating pod {%s} in namespace {%s}",
+					podName,
+					namespaceObj.Name,
+				)
+
+				By("verifying pod status as not running")
+				podPhase := ops.GetPodStatusEventually(createdPod)
+				Expect(podPhase).ToNot(Equal(corev1.PodRunning), "while verifying pod running status")
```
mynktl: nit: can we add a check to validate the reason for not running state?

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+				By("verifying pod count as 0")
+				podCount := ops.GetPodCountEventually(namespaceObj.Name, label, nil, 0)
+				Expect(podCount).To(Equal(0), "while verifying pod count")
+
+				By("deleting the PVC with StorageClass " + scName)
+				pvName := ops.GetPVNameFromPVCName(pvcName)
+				Expect(pvName).To(
+					BeEmpty(),
+					"while getting Spec.VolumeName from "+
+						"PVC {%s} in namespace {%s}",
```
mynktl:
```suggestion
					"while getting Spec.VolumeName from PVC {%s} in namespace {%s}",
```

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+	"math/rand"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const (
+	imageDirectory = "/tmp"
```
mittachaitu: nit: We can use ioutil.TempDir("/tmp","") to do all disk related stuff... This can help to run test in parallel if applicable

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+	}
+	return nil
+}
+
+func (disk *Disk) createDiskImage() error {
+	// no of blocks
+	/*count := disk.Size / blockSize
+	createImageCommand := "dd if=/dev/zero of=" + disk.imageName + " bs=" + strconv.Itoa(blockSize) + " count=" + strconv.Itoa(int(count))
+	err := utils.RunCommand(createImageCommand)*/
```
mittachaitu: We can remove this commented code

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+		"while formatting the disk {%+v} with xfs fs",
+		physicalDisk,
+	)
+
+	fmt.Println("Calling mkdir function")
+	err = disk.MkdirAll(hostPath)
+	Expect(err).To(
+		BeNil(),
+		"while making a new directory {%s}",
+		xfsHostpathDir,
```
mynktl: It should be `hostPath`. We can remove `xfsHostpathDir` since it is being used in error message only

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return fmt.Errorf("error truncating disk image. Error : %v", err)
+	}
+
+	return nil
+}
+
+func (disk *Disk) createLoopDevice() error {
+	var err error
+	if _, err = os.Stat(disk.imageName); err != nil {
```
mittachaitu: Can we createDiskImage only if the error is not exist?
```suggestion
	if _, err = os.Stat(disk.imageName); err != nil && os.IsNotExist(err) {
	 // Create sparse file
	} else if err != nil {
	   return err
	}
```

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+	return RunCommand("sudo " + cmd)
+}
+
+// RunCommand runs a command on the host
+func RunCommand(cmd string) error {
+	substring := strings.Fields(cmd)
+	name := substring[0]
+	args := substring[1:]
+	stdout, err := exec.Command(name, args...).CombinedOutput() // #nosec G204
+	fmt.Printf("%+v \n", string(stdout))
```
mittachaitu: We can remove this debug output

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return fmt.Errorf("run failed %s %v", cmd, err)
```
mittachaitu: Let's also log stdout(in fmt.Errorf) since err will only contain the `exit code` of the command that it got executed

tests/disk/disk.go
```diff
@@ -0,0 +1,208 @@
...
+	deleteBackingImageCommand := "rm " + disk.imageName
+	err = RunCommandWithSudo(deleteBackingImageCommand)
+	if err != nil {
+		return fmt.Errorf("could not delete backing disk image. Error : %v", err)
+	}
+	deleteLoopDeviceCommand := "rm " + disk.Name
+	err = RunCommandWithSudo(deleteLoopDeviceCommand)
+	if err != nil {
+		return fmt.Errorf("could not delete loop device. Error : %v", err)
+	}
```
mittachaitu: Can we use os.RemoveAll/os.Remove/os.RemoveDirectory to delete entries?

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+	xfsHostpathDir := "/var/openebs/integration-test/xfs/"
+	physicalDisk := disk.NewDisk(DiskImageSize)
+
+	fmt.Println("Calling attach function")
+	err := physicalDisk.AttachDisk()
+	Expect(err).To(
+		BeNil(),
+		"while creating loopback device with disk {%+v}",
+		physicalDisk,
+	)
```
mittachaitu: As a best practice let's not use gomega assertions in functions... Let's return error if there is any error and if caller is for test then we can have assertions over there... 
```suggestion
	if err != nil {
	    return errors.Wrapf(err, "while creating loop back device with disk %+v", physicalDisk)
	}
```

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+
+	fmt.Println("Calling fs function")
```
mittachaitu: We can remove debuging prints

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+		physicalDisk,
+	)
+
+	fmt.Println("Calling mkdir function")
+	err = disk.MkdirAll(hostPath)
+	Expect(err).To(
+		BeNil(),
+		"while making a new directory {%s}",
+		xfsHostpathDir,
+	)
```
mittachaitu: We are creating an entry of hostPath but logging xfsHostPathDir which leads to issue during debug time if any error occurs

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+		physicalDisk,
+	)
+
+	// Detach and delete the disk
+	err = physicalDisk.DetachAndDeleteDisk()
+	Expect(err).To(
+		BeNil(),
+		"while detaching and deleting the disk {%+v}",
+		physicalDisk,
+	)
```
mittachaitu: Let's also delete the directory that we created during `PrepareDisk`?

tests/operations.go
```diff
@@ -230,6 +230,16 @@ func (ops *Operations) withDefaults() {
 	}
 }
 
+// GetPodStatusEventually gives the phase of the pod eventually
+func (ops *Operations) GetPodStatusEventually(pod *corev1.Pod) corev1.PodPhase {
+	var podStatus corev1.PodPhase
+	for i := 0; i < maxRetry; i++ {
+		podStatus = pod.Status.Phase
+		time.Sleep(5 * time.Second)
```
mynktl, line 242: we are returning the status value from the given argument. we don't need to add delay for that.

tests/operations.go
```diff
@@ -230,6 +230,16 @@ func (ops *Operations) withDefaults() {
...
+	}
```
mittachaitu: For what pod phase we are waiting for?

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+							WithName("busybox").
+							WithImage("busybox").
+							WithCommandNew(
+								[]string{
+									"/bin/sh",
+									"-c",
+								},
+							).
+							WithArgumentsNew([]string{
+								"sleep 150000;",
```
mynktl: why sleep here?

mynktl (changes requested):

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+								"for i in $(seq 1 3); do sudo dd if=/dev/zero of=" + "pvc-" + string(newPvc.UID) + "/test.txt bs=1M count=$i 2>/dev/null; if [ $? -ne 0 ]; then echo \"filesize of $i Mb failed, quota reached!\"; exit 1; else echo \"filesize of $i Mb was ok\"; fi; done",
+							}).
```
mittachaitu: With this, we can't ensure whether capacity is reached or not deploy busy box app and let's [exec](https://github.com/openebs/dynamic-nfs-provisioner/blob/develop/tests/exec.go#L31) into the pod and run the command

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+							WithImage("busybox").
+							WithCommandNew(
+								[]string{
+									"/bin/sh",
+									"-c",
+								},
+							).
+							WithArgumentsNew([]string{
+								"sleep 150000;",
+								"for i in $(seq 1 3); do sudo dd if=/dev/zero of=" + "pvc-" + string(newPvc.UID) + "/test.txt bs=1M count=$i 2>/dev/null; if [ $? -ne 0 ]; then echo \"filesize of $i Mb failed, quota reached!\"; exit 1; else echo \"filesize of $i Mb was ok\"; fi; done",
```
mittachaitu: We should write at dir /mnt/store to verify for the same

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+			volume.NewBuilder().
+				WithName("demo-vol1").
+				WithPVCSource(pvcName),
+		).
+		Build()
+}
+
+// PrepareDisk prepares the setup necessary for testing xfs hostpath quota
+func PrepareDisk(fsType, hostPath string) disk.Disk {
+	xfsHostpathDir := "/var/openebs/integration-test/xfs/"
```
mittachaitu: We can remove this variable... since we are using hostPath

mittachaitu (commented):
provided few comments PTAL

pkg/kubernetes/api/core/v1/pod/build.go
Ab-hishek: This is a used to build a new pod with selectors. And a new pod won't have labels at all present before since it is being newly formed.

tests/disk/disk.go
Ab-hishek: We don't want the actual test to start or run in parallel before the disk gets prepared.

tests/disk/disk.go
Ab-hishek: We need sudo to remove all the entries. And os package functions can't perform this deletion task without sudo access.

.github/workflows/pull_request.yml
```diff
@@ -131,6 +131,7 @@ jobs:
 
       - name: Set up infra for integration test
         run: |
+          sudo apt-get update && sudo apt-get install xfsprogs
```
niladrih: This line is not required, as 'xfsprogs' package is pre-installed in the runner.
You may leave it here for the sake of redundancy.... in that case you will have to include the '-y' flag to the install command to bypass the interactive (Y/N) question.

Makefile
niladrih: Why is this timeout required? Can we handle this internally with sleep in the step which is facing timeout issues?

tests/disk/disk.go
niladrih: Using ioutil.TempDir or mktemp is a better option here.

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+	sc "github.com/openebs/dynamic-localpv-provisioner/pkg/kubernetes/api/storage/v1/storageclass"
+	"github.com/openebs/dynamic-localpv-provisioner/tests/disk"
+)
+
+const (
+	// DiskImageSize is the default file size(1GB) used while creating backing image
+	DiskImageSize = 1073741824
+)
+
+var _ = Describe("TEST HOSTPATH LOCAL PV", func() {
```
niladrih: This test spec has the same name as the spec in hostpath_test.go.
This should be renamed with keywords like 'XFS', 'quota', 'hostpath' included in the name. The Makefile regex can be altered to run this test spec along with the other hostpath test spec in the 'focus' and 'skip' flags.

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+		scNamePrefix   = "sc-hp-xfs"
+		scName         string
+		podObj         *corev1.Pod
+		labelselector  = map[string]string{
+			"demo": "hostpath-pod",
+		}
+	)
+	physicalDisk := disk.Disk{}
+
+	When("preparing the loopback device with xfs fs", func() {
```
niladrih: These operations should be written together with the variable declarations above as 'Before Spec' instructions. These should not be a part of the testing themselves, i.e. these intructions should not be withing When/It blocks.

When creating StorageClass/PVC, the storageClass creation/PVC creation are themselves included in the integration test, and so they may be written with the When/It blocks.

Here you may simply call PrepareDisk and handle the error with Ginkgo/Gomega assertions.

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+
+				By("verifying PVC is deleted")
+				status := ops.IsPVCDeletedEventually(pvcName, namespaceObj.Name)
+				Expect(status).To(
+					BeTrue(),
+					"when checking status of deleted PVC {%s}",
+					pvcName,
+				)
+
+				By("deleting the storageClass " + scName)
```
niladrih: StorageClass deletion is handled as a part of the AfterSuite cleanup, explicit cleanup is not required.
Check suite_test.go file for more context.

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+	physicalDisk := disk.Disk{}
+
+	When("preparing the loopback device with xfs fs", func() {
+		It("should create and mount the disk", func() {
+			physicalDisk = PrepareDisk("xfs", xfsHostpathDir)
+		})
+	})
+
+	When("hostpath is not having xfs filesystem", func() {
+		When("pod consuming pvc with a valid quota storageclass is created", func() {
```
niladrih: This way of embedding one When block in another is affecting readability.
You may use a 'Context' block here.
You should consider splitting the tests into separate specs, if the test cases are too varied.

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,606 @@
...
+				Expect(err).To(
+					BeNil(),
+					"while deleting storageclass {%s}",
+					scName,
+				)
+			})
+		})
+
+		When("pod consuming pvc with quota applied created", func() {
+			It("should be up and running and should not be able to write more than the quota limit", func() {
```
niladrih: The scope of this 'It' block is too large.

niladrih (changes requested):
Left few comments.

Makefile
Ab-hishek: Trying to resolve this with the internal sleep commands.

tests/hostpath_quota_test.go
Ab-hishek: Those will be deleted after en entire suite completes. I want different storage classes within the suite.

tests/operations.go
Ab-hishek: Trying to refactor it.

tests/hostpath_quota_test.go
niladrih: This is fine. You can have multiple storageClasses created at the same time. Because you are using GenerateName, they will all be named uniquely.

tests/hostpath_quota_test.go
Ab-hishek: removing this. Not required.

pkg/kubernetes/api/core/v1/pod/build.go
Ab-hishek: However changed the code to support appending labels.

tests/hostpath_quota_test.go
Ab-hishek: Changed the code logic to check for exact state expected.

tests/hostpath_quota_test.go
Ab-hishek: Calling this from outside of the When/It block doesn't create the disk. The function is being called but no action is being performed. Don't know why this is happening.

.github/workflows/pull_request.yml
```diff
@@ -131,6 +131,7 @@ jobs:
 
       - name: Set up infra for integration test
         run: |
+          sudo apt-get update && sudo apt-get install -y xfsprogs
```
niladrih, line 134: nit: Please add this change to build.yml workflow as well. Thanks.

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,463 @@
...
+	When("writing data more than quota limit into the hostpath volume", func() {
+		It("should not be able to write more than the enforced limit", func() {
+			By("Verifying the quota applied on the volume works")
+			execOption := NewOptions()
+			// The command formed below is attempting to create a file named “test.txt” inside the attached volume,
+			// its size increasing 1M in each iteration of the loop should show us the quota in action.
+			option := execOption.WithPodName(createdPod.Name).
+				WithContainer("busybox").
+				WithNamespace(namespaceObj.Name).
+				WithCommand([]string{"/bin/sh", "-c", fmt.Sprintf(`for i in $(seq 1 7); do dd if=/dev/zero of=/mnt/store1/test.txt bs=1M count=$i 2>/dev/null; if [ $? -ne 0 ]; then echo "filesize of $i Mb failed, quota reached!"; else echo "filesize of $i Mb was ok"; fi; done`)}...)
```
niladrih: Could you test if we are hitting the quota limit by using something like `du -h --summarize --block-size=M /mnt/store1` on the mountpoint?

Also why are we testing by writing in 1 MB increments, and not directly trying to exceed the quota limit? The error message returned by 'dd' could be useful if printed to make sure that the error was due to lack of capacity.

Can we do something like this...
```bash
dd if=/dev/zero of=/mnt/store1/test.txt bs=1M count=100 status=none || du -hs -BM /mnt/store1
```
This prints nothing if write succeeds, and does an exit 0, prints error and final size if write fails.
To add exit status value, you can do this also...
```bash
dd if=/dev/zero of=/mnt/store1/test.txt bs=1M count=100 status=none || (du -hs -BM /mnt/store1 && exit 1)
```

niladrih (changes requested):
Changes are good.
Left few comments.

tests/hostpath_quota_test.go
mittachaitu: > Also why are we testing by writing in 1 MB increments, and not directly trying to exceed the quota limit? The error message returned by 'dd' could be useful if printed to make sure that the error was due to lack of capacity.

It is required to test the behavior of feature... like if we write less data it should succeed once we start consuming over than limit it will throw error

pkg/kubernetes/api/core/v1/pod/build.go
```diff
@@ -90,6 +91,34 @@ func (b *Builder) WithNamespace(namespace string) *Builder {
...
+		)
+		return b
+	}
+
+	// copy of original map
+	newmatchlabels := map[string]string{}
+	for key, value := range matchlabels {
+		newmatchlabels[key] = value
+	}
+
```
mittachaitu, line 107: For appending we also need to add below code 
```suggestion
for key, value := range b.pod.object.Labels {
    newmatchlabels[key] = value
}
```

tests/disk/disk.go
```diff
@@ -0,0 +1,196 @@
...
+// CreateLoopDevice creates a loopback device if the disk is not present
+func (disk *Disk) CreateLoopDevice() error {
+	if disk.DiskPath == "" {
+		var err error
+		if _, err = os.Stat(disk.imageName); err != nil && os.IsNotExist(err) {
+			err = disk.createDiskImage()
+			if err != nil {
+				return err
+			}
+		}
```
mittachaitu, line 96: Missing below else part else original error will be swallowed
```suggestion
		} else if err != nil {
		     return err
		}
```

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,463 @@
...
+
+			By("verifying PVC is deleted")
+			status := ops.IsPVCDeletedEventually(pvcName, namespaceObj.Name)
+			Expect(status).To(
+				BeTrue(),
+				"when checking status of deleted PVC {%s}",
+				pvcName,
+			)
+		})
+	})
```
mittachaitu, line 189: We also need to delete the StorageClass created in step1

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,463 @@
...
```
mittachaitu, line 361: Here, also StorageClass created for test needs to be deleted

tests/operations.go
```diff
@@ -635,19 +661,24 @@ func (ops *Operations) ExecPod(opts *Options) ([]byte, error) {
...
 	err = exec.Stream(remotecommand.StreamOptions{
 		Stdout: &execOut,
 		Stderr: &execErr,
 		Tty:    false,
 	})
-	Expect(err).To(BeNil(), "while streaming the command in pod ", opts.podName, execOut.String(), execErr.String())
+	/*Expect(err).To(BeNil(), "while streaming the command in pod ", opts.podName, execOut.String(), execErr.String())
 	Expect(execOut.Len()).Should(BeNumerically(">=", 0), "while streaming the command in pod ", opts.podName, execErr.String(), execOut.String())
-	return execOut.Bytes(), nil
+	return execOut.Bytes(), nil*/
```
mittachaitu: nit: We can remove commented code

mittachaitu (commented):
test looks good... we are leaving resources which we created for test good to delete those resources

pkg/kubernetes/api/core/v1/pod/build.go
Ab-hishek: Ahh!! missed it.

tests/hostpath_quota_test.go
Ab-hishek: That is handled in AfterSuite block.

tests/hostpath_quota_test.go
Ab-hishek: same comment as above.

pkg/kubernetes/api/core/v1/pod/build.go
mynktl: you don't need to use an additional map. you can directly add to the existing pod label..

tests/disk/disk.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// RunCommand runs a command on the host
+func RunCommand(cmd string) error {
+	substring := strings.Fields(cmd)
+	name := substring[0]
+	args := substring[1:]
+	stdout, err := exec.Command(name, args...).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("run failed %s %v %v", cmd, err, stdout)
```
mynktl: To identify the value in logging
```suggestion
		return fmt.Errorf("run failed, cmd={%s} error={%v} output={%v}", cmd, err, stdout)
```

tests/hostpath_quota_test.go
niladrih: Check criteria might not be sufficient here.

With a check like `dd if=/dev/zero of=/mnt/store1/test.txt bs=1M count=100 || du -hs -BM /mnt/store1`, whatever no. of bytes were written successfully, is printed --> proof of successful write prior to write failure.

Not sure if the looping incremental writing test is scalable for testing with a larger data set, and every write begins from the same offset resulting in redundant writing,

tests/disk/disk.go
```diff
@@ -0,0 +1,197 @@
...
+	disk.MountPoints = append(disk.MountPoints, path)
+	return nil
+}
+
+func (disk *Disk) Unmount() error {
+	var lastErr error = nil
+	for i := 0; i < len(disk.MountPoints); i++ {
+		err := RunCommandWithSudo("umount " + disk.MountPoints[i])
+		if err != nil {
+			lastErr = err
```
mynktl: lets append the error and return all the errors

tests/hostpath_quota_test.go
mittachaitu: Creating at one place and deleting at a different places is not a good thing... But it is ok if current test code is doing the same

mittachaitu (approved):
LGTM

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,463 @@
...
+	When("StorageClass with valid xfs quota parameters is created", func() {
+		It("should create a StorageClass", func() {
+			By("building a StorageClass")
+			scObj, err = sc.NewStorageClass(
+				sc.WithGenerateName(scNamePrefix),
+				sc.WithLabels(map[string]string{
+					"openebs.io/test-sc": "true",
+				}),
+				sc.WithLocalPV(),
+				sc.WithHostpath("/var/openebs/integration-test"),
```
mynktl: `/var/openebs/integration-test` is being used at other places also. lets use a constant for this value

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,463 @@
...
+				"while creating StorageClass with name prefix {%s}",
+				scNamePrefix,
+			)
+			scName = scObj.ObjectMeta.Name
+		})
+	})
+
+	When("pvc with storageclass "+scName+" is created", func() {
+		It("should create a pvc", func() {
+			By("building a PVC with StorageClass " + scName)
```
mynktl, line 93: `scName` is being set by the first `When` container block. Better to verify if it is empty or not. Add this to other blocks also wherever we are using values from the different blocks. You can use
```golang
Expect(scName).NotTo(BeEmpty(), "SC name should not be empty")
```

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,463 @@
...
+		WithVolumeBuilder(
+			volume.NewBuilder().
+				WithName("demo-vol1").
+				WithPVCSource(pvcName),
+		).
+		Build()
+}
+
+// PrepareDisk prepares the setup necessary for testing xfs hostpath quota
+func PrepareDisk(fsType, hostPath string) (disk.Disk, error) {
```
mynktl: `PrepareDisk` and `DestroyDisk` can be moved to tests/disk

tests/operations.go
```diff
@@ -230,6 +229,18 @@ func (ops *Operations) withDefaults() {
 	}
 }
 
+// CheckPodStatusEventually gives the phase of the pod eventually
+func (ops *Operations) CheckPodStatusEventually(pod *corev1.Pod, expectedPodPhase corev1.PodPhase) bool {
+	//var podStatus corev1.PodPhase
+	for i := 0; i < maxRetry; i++ {
```
mynktl, line 234: you need to fetch the latest pod object and check the status against `expectedPodPhase`

tests/hostpath_quota_test.go
```diff
@@ -0,0 +1,463 @@
...
+			execOption := NewOptions()
+			// The command formed below is attempting to create a file named “test.txt” inside the attached volume,
+			// its size increasing 1M in each iteration of the loop should show us the quota in action.
+			option := execOption.WithPodName(createdPod.Name).
+				WithContainer("busybox").
+				WithNamespace(namespaceObj.Name).
+				WithCommand([]string{"/bin/sh", "-c", fmt.Sprintf(`for i in $(seq 1 7); do dd if=/dev/zero of=/mnt/store1/test.txt bs=1M count=$i 2>/dev/null; if [ $? -ne 0 ]; then echo "filesize of $i Mb failed, quota reached!"; else echo "filesize of $i Mb was ok"; fi; done`)}...)
+			stdOut, stdErr, err := ops.ExecPod(option)
+			fmt.Printf("When running command to test enforced quota. stdOut: {%s}, stderr: {%s}, error: {%v}", stdOut, stdErr, err)
+			Expect(stdErr).NotTo(BeNil(), "trying to write beyond the quota limit should not be allowed")
```
mynktl: need to check if `err` is `nil` or not

tests/operations.go
Ab-hishek: pod object whose status I need to check is passed as a parameter.

tests/disk/disk.go
```diff
@@ -0,0 +1,260 @@
...
+	err = physicalDisk.Mount(hostPath)
+	if err != nil {
+		return physicalDisk, errors.Wrapf(err, "while mounting the disk with pquota option {%+v}", physicalDisk)
+	}
+
+	return physicalDisk, nil
+}
+
+// DestroyDisk performs performs the clean-up task after testing the features
+func DestroyDisk(physicalDisk Disk, hostPath string) error {
```
mynktl: we can bind this fn to *Disk

mynktl (approved):

tests/hostpath_quota_test.go
niladrih:
```suggestion
				WithCommand([]string{"/bin/sh", "-c", "dd if=/dev/zero of=/mnt/store1/test.txt bs=1M count=10 2>&1 || du -sm /mnt/store1 | cut -f -1 | tr -d '\n'1>&2"}...)
```

The above command can be used to try writing 10M. And the expected result is that the command should fail --> this being a -ve test.

Changes for the variable initialization part -- 
```golang
capacity = "5M"
```
to
```golang
capacityInMib = "5"
capacity = capacityInMib + "M"
```

The assertions after evaluating stdout and stderr can be something like this
```golang
Expect(stdOut).NotTo(BeEmpty(), "trying to write till the quota limit should be allowed")
Expect(stdErr).To(Equal(capacityInMib), "trying to write beyond the quota limit should not be allowed")
```

niladrih (approved):
LGTM

PR #100, opened 2021-09-10
chore(integration-test): fix flakiness

niladrih:

The integration tests are used to test new changes to the codebase. They should produce reliable results when run.

**What this PR does?**:
Changes the value of the maxRetry variable in tests/operations.go file. This extends the timeout duration for checks to a value greater than or equal to 5 minutes.

Using ginkgo cli, instead of go test to work around the default timeout at the 10m mark
Updated tests/README.md
Updated Minikube version and Kubernetes version for Minikube in CI.

No.

N/A

No.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,13 +182,13 @@ jobs:
       - name: Set up Minikube
         uses: manusa/actions-setup-minikube@v2.4.2
         with:
-          minikube version: 'v1.22.0'
-          kubernetes version: 'v1.22.0'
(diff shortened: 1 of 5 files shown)


Ab-hishek:
This won't fix the issue. Issue is with test timing out. We will have to increase the default test time which 10min. The entire suite is taking more than that. That is why the tests are failing. Check this out -> https://github.com/golang/go/issues/25886,

niladrih:
I think I'm going to move back to the ginkgo cli

kmova (approved):

PR #107, opened 2021-09-15
refact(XFS-Quota): get XFS Quota parameters from StorageClass 'cas.openebs.io/config' annotation

niladrih:
## Pull Request template

This PR makes code changes for XFS Quota so that it may use cas-config parameters instead of spec.Parameters.

**What this PR does?**:
Modifies volumeConfig struct to include a new map generated for the 'Data' field in Config object.
Adds functions to get values of Enabled and Data fields.
Adds a function and modifies existing code to use values from StorageClass's cas-config annotation.
Sample storageClass:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-hostpath-xfs
  annotations:
    openebs.io/cas-type: local
    cas.openebs.io/config: |
      - name: StorageType
        value: "hostpath"
      - name: BasePath
        value: "/var/openebs/local/"
      - name: XFSQuota
        enabled: "true"
        data:
          softLimitGrace: "80%"
          hardLimitGrace: "85%"
provisioner: openebs.io/local
volumeBindingMode: WaitForFirstConsumer
reclaimPolicy: Delete
```
This above may be used in place of the below StorageClass with similar configuration:
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-hostpath-xfs
  annotations:
    openebs.io/cas-type: local
    cas.openebs.io/config: |
      - name: StorageType
        value: "hostpath"
      - name: BasePath
        value: "/var/openebs/local/"
provisioner: openebs.io/local
volumeBindingMode: WaitForFirstConsumer
reclaimPolicy: Delete
parameters:
  enableXfsQuota: 'true'
  softLimitGrace: "80%"
  hardLimitGrace: "85%"
```
-----------------------------------------

Sample StorageClass for when you want to enforce the PVC's `.spec.resources.requests.storage` value.
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-hostpath-xfs
  annotations:
    openebs.io/cas-type: local
    cas.openebs.io/config: |
      - name: StorageType
        value: "hostpath"
      - name: BasePath
        value: "/var/openebs/local/"
      - name: XFSQuota
        enabled: "true"
provisioner: openebs.io/local
volumeBindingMode: WaitForFirstConsumer
reclaimPolicy: Delete
```

### DIFF
--- a/cmd/provisioner-localpv/app/config.go
+++ b/cmd/provisioner-localpv/app/config.go
@@ -19,6 +19,7 @@ package app
 
 import (
 	"context"
+	"strconv"
 	"strings"
 
 	mconfig "github.com/openebs/maya/pkg/apis/openebs.io/v1alpha1"
@@ -113,6 +114,17 @@ const (
 	// is specified with PVC and is useful for granting shared access
 	// to underlying hostpaths across multiple pods.
 	//KeyPVAbsolutePath = "AbsolutePath"
+
+	//KeyXFSQuota enables/sets parameters for XFS Quota.
+	// Example StorageClass snippet:
+	//    - name: XFSQuota
+	//      enabled: true
+	//      data:
+	//        softLimitGrace: "80%"
+	//        hardLimitGrace: "85%"
+	KeyXFSQuota          = "XFSQuota"
+	KeyXfsQuotaSoftLimit = "softLimitGrace"
+	KeyXfsQuotaHardLimit = "hardLimitGrace"
 )
 
 const (
@@ -168,11 +180,17 @@ func (p *Provisioner) GetVolumeConfig(ctx context.Context, pvName string, pvc *c
 		return nil, errors.Wrapf(err, "unable to read volume config: pvc {%v}", pvc.ObjectMeta.Name)
 	}
 
+	dataPvConfigMap, err := dataConfigToMap(pvConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to read volume config: pvc {%v}", pvc.ObjectMeta.Name)
+	}
+
 	c := &VolumeConfig{
-		pvName:  pvName,
-		pvcName: pvc.ObjectMeta.Name,
-		scName:  *scName,
-		options: pvConfigMap,
+		pvName:     pvName,
+		pvcName:    pvc.ObjectMeta.Name,
+		scName:     *scName,
+		options:    pvConfigMap,
+		configData: dataPvConfigMap,
 	}
 	return c, nil
 }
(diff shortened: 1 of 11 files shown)


cmd/provisioner-localpv/app/config.go
```diff
@@ -262,6 +280,32 @@ func (c *VolumeConfig) GetPath() (string, error) {
...
+	//     i. The value was something other than a straightforward
+	//        true or false
+	//    ii. The value was empty
+	if err != nil {
+		softLimit := strings.TrimSpace(c.getData(KeyXFSQuota, KeyXfsQuotaSoftLimit))
+		hardLimit := strings.TrimSpace(c.getData(KeyXFSQuota, KeyXfsQuotaHardLimit))
+		// Checking if XFS enabling is implied
+		// by the usage of 'KeyXfsQuotaSoftLimit' or
+		// 'KeyXfsQuotaHardLimit'
+		if len(softLimit) > 0 || len(hardLimit) > 0 {
```
kmova: Is this check still required?

pkg/kubernetes/api/storage/v1/storageclass/build.go
```diff
@@ -186,6 +191,41 @@ func WithDevice() StorageClassOption {
...
 
+func WithXfsQuota(softLimit, hardLimit string) StorageClassOption {
+	return func(s *storagev1.StorageClass) error {
+		if !isCompatibleWithXfsQuota(s) {
+			return errors.New("Failed to set XFSQuota parameters. " +
+				"Invalid existing '" + string(mconfig.CASConfigKey) + "' annotation" +
+				" parameters or Provisioner name.")
+		}
+
+		config := "- name: XFSQuota\n" +
```
kmova, line 204: We can probably get back to this and refactor a bit. 

Can you add a todo comment here?

cmd/provisioner-localpv/app/config.go
niladrih: In case the Data map is nil, we get an empty string. 

Inside this `if err != nil` block, we don't have a straightforward answer if we want to enable XFS Quota.
The default for XFS Quota is ----> disabled state.

But if someone skips the enabled flag, or messes it up, but has put in Limits, then we can for sure assume he has intentions of enabling XFS Quota.

If the length of the limits are not > 0 AND the 'enabled' value is not straightfowardly 'true' (inside the if err != nil block, there is not straightforward answer). XFS Quota should remain disabled.

Only described limits and explicity enabled: true should enable it.

What do you make of this @kmova ?

niladrih: > Is this check still required?

The default behavior of 0%, 0% will be triggered further down the stack. At this point, empty limits are empty limits.

pkg/kubernetes/api/storage/v1/storageclass/build.go
niladrih: Yep.

cmd/provisioner-localpv/app/provisioner_hostpath.go
```diff
@@ -93,11 +93,9 @@ func (p *Provisioner) ProvisionHostPath(ctx context.Context, opts pvController.P
...
 	}
 
-	enableXfsQuota := opts.StorageClass.Parameters[EnableXfsQuota]
-
-	if enableXfsQuota == "true" {
-		softLimitGrace := opts.StorageClass.Parameters[SoftLimitGrace]
-		hardLimitGrace := opts.StorageClass.Parameters[HardLimitGrace]
+	if volumeConfig.IsXfsQuotaEnabled() {
+		softLimitGrace := volumeConfig.getData(KeyXFSQuota, KeyXfsQuotaSoftLimit)
+		hardLimitGrace := volumeConfig.getData(KeyXFSQuota, KeyXfsQuotaHardLimit)
```
kmova, line 98: After getting these values - should we call `isValidXfsQuotaData` ?

cmd/provisioner-localpv/app/config.go
niladrih: That function is there to judge from the StorageClass config, whether to go into XFS Quota or not. That's all.

niladrih: Disabled implied enabling of XFS Quota. Explicit enable is required to use it now.

kmova (approved):

Ab-hishek (approved):
looks good. Thanks @niladrih for the refactoring.

PR #119, opened 2021-10-08
refact(integration-test): fix integration test

niladrih:

The integration tests for this repo fail due to incorrect checking criteria. This PR refines checking criteria and similar changes.

**What this PR does?**:
Changes summary:
1. Checks for BDC delete (in case of HOSTDEVICE tests) before moving on to next set of tests. This fixes an issue where a dangling BDC will claim the lone available BD and disrupt testing conditions.
2. Checks if NDM pods are running and a single BD with-no-filesystem is present before creating local-device volumes.
3. Checks for BD cleanup (in case of HOSTDEVICE tests) before moving on to next set of tests. This fixes an issue introduced by the above check (point no. 2). BD's which were not cleaned up generate a false negative when checking for the availability of 1 single no-filesystem BD.
                i. Adds utility methods for listing and filtering corev1.Events.
4.  If NDM components are present, the loop device created for XFS Quota tests is added to the NDM config path-filter's exclude list. This avoids detecting the loop device as a usable BD with NDM's ChangeDetection feature (in case /dev/loop devices are removed from exclude list).
                i. Adds utility methods for NDM Config.
5. The XFS Quota tests require 'sudo' to create and delete loop devices and other privileged operations (mkfs, mount). The sudo password prompt is issued 2 times during the test, once during BeforeSuite() and once during AfterSuite(). This is inconvenient. Added sudo as a prerequisite for the test suite.
                i. Updates tests/README.md and Makefile with sudo enabled command.
6. The XFS Quota tests assumes the directory used for HOSTPATH LOCAL PV tests to have a non-XFS filesystem. This invalidates the tests results. Made changes so that the loop device is formatted with ext4 filesystem and mounted.
                i. Adds utility function to run wipefs.
                ii. Adds mkfs option for ext4 filesystem
                iii. Adds mount option for ext4 filesystem (with Prjquota)
7. Mounts ext4 filesystem for NON-XFS FILESYSTEM xfs quota test with Prjquota to validate that non-xfs quota is not supported.
8. Modify [-ve] tests for HOSTDEVICE and XFS QUOTA Pod-count check criteria to check for non-negative result, instead of positive result.
9. Changes to directory structure used for XFS QUOTA tests -- moved disk image directory to primary integration test directory.
                i. Directory names are now generated to not stuff tests files with stale/existing test files.
10. Moved BuildPod() and BuildPersistentVolumeClaim() methods (used in XFS QUOTA tests) from the main test file to operations.go file.
11. Removed unused test artefacts which were left behind after the tests were copied over from openebs/maya.
12. Made changes to GitHub Actions CI to accommodate test prerequisites.

No.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -190,12 +190,14 @@ jobs:
       - name: Set up Minikube
         uses: manusa/actions-setup-minikube@v2.4.2
         with:
-          minikube version: 'v1.23.0'
-          kubernetes version: 'v1.22.1'
+          minikube version: 'v1.23.2'
+          kubernetes version: 'v1.22.2'
           driver: none
+          start args: '--install-addons=false'
 
       - name: Set up infra for integration test
         run: |
+          sudo apt-get update && sudo apt-get install -y xfsprogs quota linux-modules-extra-azure
           go install github.com/onsi/ginkgo/ginkgo@v1.16.4
           sudo fallocate -l 2.5GiB /mnt/disk.img
           sed -e "/path-filter/{N;N;N;s|\"\"|\"$(sudo losetup -fP /mnt/disk.img --show)\"|}" -e '/path-filter/{N;N;N;N;s|/dev/loop,||}' -e '/openebs-provisioner-hostpath/{N;s/IfNotPresent/Never/}' deploy/kubectl/openebs-operator-lite.yaml | kubectl apply -f -
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -191,7 +191,7 @@ jobs:
         uses: manusa/actions-setup-minikube@v2.4.2
         with:
           minikube version: 'v1.23.2'
-          kubernetes version: 'v1.22.2'
+          kubernetes version: 'v1.22.3'
           driver: none
           start args: '--install-addons=false'
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -197,7 +197,7 @@ jobs:
 
       - name: Set up infra for integration test
         run: |
-          sudo apt-get update && sudo apt-get install -y xfsprogs quota linux-modules-extra-azure
+          sudo apt-get update && sudo apt-get install -y xfsprogs quota linux-modules-extra-`uname -r`
           go install github.com/onsi/ginkgo/ginkgo@v1.16.4
           sudo fallocate -l 2.5GiB /mnt/disk.img
           sed -e "/path-filter/{N;N;N;s|\"\"|\"$(sudo losetup -fP /mnt/disk.img --show)\"|}" -e '/path-filter/{N;N;N;N;s|/dev/loop,||}' -e '/openebs-provisioner-hostpath/{N;s/IfNotPresent/Never/}' deploy/kubectl/openebs-operator-lite.yaml | kubectl apply -f -
--- a/pkg/kubernetes/ndmconfig/ndmconfig.go
+++ b/pkg/kubernetes/ndmconfig/ndmconfig.go
@@ -1,0 +1,162 @@
+/*
+Copyright 2021 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ndmconfig
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type Config struct {
+	ProbeConfigs  []ProbeConfig  `yaml:"probeconfigs,omitempty"`  // ProbeConfigs contains configs of Probes
+	FilterConfigs []FilterConfig `yaml:"filterconfigs,omitempty"` // FilterConfigs contains configs of Filters
+	TagConfigs    []TagConfig    `yaml:"tagconfigs,omitempty"`    // TagConfigs contains configs for tags
+}
+
+// ProbeConfig contains configs of Probe
+type ProbeConfig struct {
+	Key   string `yaml:"key"`   // Key is key for each Probe
+	Name  string `yaml:"name"`  // Name is name of Probe
+	State string `yaml:"state"` // State is state of Probe
+}
+
+// FilterConfig contains configs of Filter
+type FilterConfig struct {
+	Key     string `yaml:"key"`               // Key is key for each Filter
+	Name    string `yaml:"name"`              // Name is name of Filer
+	State   string `yaml:"state"`             // State is state of Filter
+	Include string `yaml:"include,omitempty"` // Include contains , separated values which we want to include for filter
+	Exclude string `yaml:"exclude,omitempty"` // Exclude contains , separated values which we want to exclude for filter
+}
+
+type TagConfig struct {
+	Name    string `yaml:"name,omitempty"`
+	Type    string `yaml:"type,omitempty"`
+	Pattern string `yaml:"pattern,omitempty"`
+	TagName string `yaml:"tag,omitempty"`
+}
+
+// Type of list -- include, exclude
+type ListType string
+
+const (
+	Include ListType = "include"
+	Exclude ListType = "exclude"
+)
+
+func NewConfigFromAPIConfigMap(ndmConfigMap *corev1.ConfigMap) (*Config, error) {
+	if ndmConfigMap == nil {
+		return nil, errors.New("NDM ConfigMap is 'nil'")
+	}
+
+	var c Config
+
+	ndmConfigString := ndmConfigMap.Data["node-disk-manager.config"]
+	err := yaml.Unmarshal([]byte(ndmConfigString), &c)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal NDM config")
+	}
+
+	return &c, nil
+}
+
+func (c *Config) AppendToPathFilter(listtype ListType, diskPath string) error {
+	if c == nil {
+		return errors.New("Config is nil")
+	}
+
+	var pathFilterIndex int
+	foundFlag := 0
+	for index, filterConfig := range c.FilterConfigs {
+		if filterConfig.Key == "path-filter" {
+			pathFilterIndex = index
+			foundFlag++
+			break
+		}
+	}
+	if foundFlag == 0 {
+		return errors.New("No filterconfig with 'key: path-filter' found")
+	}
+
+	separator := ""
+	switch listtype {
+	case Include:
+		if len(c.FilterConfigs[pathFilterIndex].Include) > 0 {
+			separator = ","
+		}
+		c.FilterConfigs[pathFilterIndex].Include = c.FilterConfigs[pathFilterIndex].Include + separator + diskPath
+		return nil
+	case Exclude:
+		if len(c.FilterConfigs[pathFilterIndex].Exclude) > 0 {
+			separator = ","
+		}
+		c.FilterConfigs[pathFilterIndex].Exclude = c.FilterConfigs[pathFilterIndex].Exclude + separator + diskPath
+		return nil
+	default:
+		return errors.New("invalid filterconfig path-filter list name")
+	}
+}
+
+func (c *Config) RemoveFromPathFilter(listtype ListType, diskPath string) error {
+	if c == nil {
+		return errors.New("Config is nil")
+	}
+
+	var pathFilterIndex int
+	foundFlag := 0
+	for index, filterConfig := range c.FilterConfigs {
+		if filterConfig.Key == "path-filter" {
+			pathFilterIndex = index
+			foundFlag++
+			break
+		}
+	}
+	if foundFlag == 0 {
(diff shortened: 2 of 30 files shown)


.github/workflows/build.yml
```diff
@@ -190,12 +190,14 @@ jobs:
...
-          minikube version: 'v1.23.0'
-          kubernetes version: 'v1.22.1'
+          minikube version: 'v1.23.2'
+          kubernetes version: 'v1.22.2'
           driver: none
+          start args: '--install-addons=false'
 
       - name: Set up infra for integration test
         run: |
+          sudo apt-get update && sudo apt-get install -y xfsprogs quota linux-modules-extra-azure
```
akhilerm: `linux-modules-extra-azure` Why is this module needed?

pkg/kubernetes/api/core/v1/event/buildlist_test.go
```diff
@@ -0,0 +1,131 @@
+/*
```
akhilerm: Do we actually need test for builders? Its kind of boilerplate code right?

pkg/kubernetes/ndmconfig/ndmconfig.go
```diff
@@ -0,0 +1,162 @@
...
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type Config struct {
```
akhilerm, line 27: Can this be imported from NDM instead of copy pasting the structs?

tests/README.md
```diff
@@ -6,12 +6,21 @@ Local PV Provisioner BDD tests are developed using ginkgo & gomega libraries.
 
 ### Pre-requisites
 
-- These tests are meant to be run in a single-node Kubernetes
-  cluster with one single available blockdevice (not mounted).
+- These tests are meant to be run in a single-node Kubernetes (v1.16+)
```
akhilerm, line 9: What is the minimum version supported ? Is it 1.16?

.github/workflows/build.yml
akhilerm: nvm, got it from the README file.

tests/disk/disk.go
```diff
@@ -82,95 +77,152 @@ func (disk *Disk) createDiskImage() error {
...
+			"Error: disk path is already set on Disk obj")
+	}
+
+	err := disk.createDiskImage(imgDir)
+	if err != nil {
+		return err
 	}
+
+	// create the loop device using losetup
+	createLoopDeviceCommand := "losetup -P " + "/dev/" + getLoopDevName() + " " + disk.imageName + " --show"
```
akhilerm, line 94: Is the `-P` flag necessary, Doesn't the system perform a partprobe  automatically?

tests/disk/disk.go
```diff
@@ -82,95 +77,152 @@ func (disk *Disk) createDiskImage() error {
...
 }
 
-func (disk *Disk) CreateFileSystem(fstype string) error {
-	createMkfsCommand := "mkfs -t " + fstype + " " + disk.DiskPath
-	err := RunCommandWithSudo(createMkfsCommand)
+func (disk *Disk) CreateFilesystem(fstype string) error {
+	var mkfsCommand string
+	switch fstype {
+	case "ext4":
+		mkfsCommand = "mkfs.ext4 -O quota -E quotatype=prjquota " + disk.DiskPath
```
akhilerm, line 142: q: Does ext4 also support quotas?

akhilerm (commented):
Given a few comments

.github/workflows/build.yml
niladrih: This is for the quota_v2 module. We are mounting an ext4 filesystem with prjquota to test if XFS quota fails here.

pkg/kubernetes/api/core/v1/event/buildlist_test.go
niladrih: Yep.

tests/README.md
niladrih: Yes.

tests/disk/disk.go
niladrih: >Does ext4 also support quotas?

Yes. 
This is to test for a negative case... because dynamic-localpv does not support ext4 quotas

tests/disk/disk.go
niladrih: Checking this one....

pkg/kubernetes/ndmconfig/ndmconfig.go
niladrih: Yes!

tests/disk/disk.go
niladrih: The present test-case just mounts the integration test directory. Hoping that the directory is ext4.
But now we make sure that the directory is ext4, and also mount it with prjquota. Giving the directory a fighting chance :P !

pkg/kubernetes/ndmconfig/ndmconfig.go
niladrih: @akhilerm -- The structs here contain tags for marshalling as well. Without the 'omitempty', the config yaml won't look clean and concise.

Checked. It is not possible to get the same result without explicitly creating a new type. Cannot use github.com/openebs/node-disk-manager/cmd/ndm_daemonset/controller.NodeDiskManagerConfig in place of Config.

pkg/kubernetes/api/core/v1/event/buildlist_test.go
niladrih: Removed buildlist_test.go

tests/disk/disk.go
niladrih: I see no reason to use the `-P` flag for my Ubuntu 20.04 system.
Found this issue https://stackoverflow.com/questions/37227233/having-losetup-read-the-partition-table
This suggests that this behavior might not be consistent across platforms. Leaving this flag for the sake of redundancy.

tests/disk/disk.go
```diff
@@ -118,59 +115,113 @@ func getLoopDevName() string {
...
+	// List partitions
+	fdiskCommand := "fdisk -o Device -l " + disk.DiskPath
+	fdiskStdOut, err := RunCommand(fdiskCommand)
 	if err != nil {
-		return fmt.Errorf("error creating hostath directory. Error : %v", err)
+		return fmt.Errorf("error listing disk partitions. Error: %v", err)
+	}
+
+	// Wipe partitions
+	stringsAndSubstrings := regexp.MustCompile("\n("+disk.DiskPath+".*)").
```
akhilerm, line 166: Why is `\n` used in the regex?

tests/disk/disk.go
```diff
@@ -196,44 +247,45 @@ func (disk *Disk) DetachAndDeleteDisk() error {
...
+		if err != nil {
+			return physicalDisk, errors.Wrapf(err, "while formatting the disk {%+v} with xfs fs", physicalDisk)
+		}
 
+		// Mount the xfs formatted loopback device
+		err = physicalDisk.PrjquotaMount(hostPath)
+		if err != nil {
+			return physicalDisk, errors.Wrapf(err, "while mounting the disk with pquota option {%+v}", physicalDisk)
+		}
+	*/
```
akhilerm: Can be removed if this section is not used

akhilerm (commented):
a few more questions @niladrih

tests/disk/disk.go
niladrih: The string we are looking for the partition paths from the following output
```bash
Disk /dev/loop13: 2.51 GiB, 2684354560 bytes, 5242880 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: CF0671E8-2143-4A62-AEF1-3F3ACB0572BA

Device
/dev/loop13p1
/dev/loop13p2
```
This may select '/dev/loop13:' from the first line as well. We don't want that.
We are relying on this output format and only looking among the substrings that start with a newline.

tests/disk/disk.go
niladrih: Removed the commented section.

tests/disk/disk.go
niladrih: Also the entire string is not subdivided line-wise like it is with 'grep'. So ^/dev/loop13 won't work, as it is in the middle, and it won't find it at the start of the string.

This implementation is trying to mimic [this](https://github.com/openebs/node-disk-manager/blob/develop/pkg/cleaner/jobcontroller.go#L107).

niladrih: The regex will find... "\n/dev/loop13p1"
It will try to find substring "/dev/loop13" withing "\n/dev/loop13"
Then it will create a slice of string slices, where each string slice will have something like "\n/dev/loop13p\<partition-number\>" as the first element and the rest will be "/dev/loop13p\<partition-number\>" (hence the reason that there can only be 1 substring)

We will use the substring in a wipefs command. We won't actually use the matched string.

tests/hostpath_quota_test.go
```diff
@@ -156,15 +179,6 @@ var _ = Describe("TEST HOSTPATH XFS QUOTA LOCAL PV WITH NON-XFS FILESYSTEM", fun
...
 	When("pvc with storageclass "+scName+" is deleted", func() {
 		It("should delete the pvc", func() {
-			By("deleting the PVC with StorageClass " + scName)
-			pvName := ops.GetPVNameFromPVCName(pvcName)
-			Expect(pvName).To(
-				BeEmpty(),
-				"while getting Spec.VolumeName from PVC {%s} in namespace {%s}",
-				pvcName,
-				namespaceObj.Name,
-			)
```
Ab-hishek: Why do we need to remove these?

niladrih: What this is doing is... checking if the PV name is empty.
This is not related to the PVC deletion context.

This check has been moved to line 152 of the same file. This check is performed when checking for other signs of failed provisioning -- Pod status, PVC status.

Ab-hishek: We can check here that the PV name to be empty where the pv name is fetched from inside the pvc object opposite of what is done everywhere else where we delete the pvc.

niladrih: Okay, I see... you are substituting the PV delete check with this.
Makes sense... put it back where it was.

Ab-hishek (approved):
lgtm.

tests/operations.go
```diff
@@ -528,6 +489,75 @@ func isNotFound(err error) bool {
 	}
 }
 
+// IsBdCleanedUpEventually tries to get the deleted BDC
+// and returns true if BDC is not found
+// else returns false
+func (ops *Operations) IsBdCleanedUpEventually(namespace, bdName, bdcName string) bool {
```
niladrih, line 495: @kmova -- Require your review on this.

tests/disk/disk.go
```diff
@@ -30,7 +32,8 @@ import (
 
 const (
 	// DiskImageSize is the default file size(1GB) used while creating backing image
-	DiskImageSize = 1073741824
+	DiskImageSize       = 1073741824
+	DiskImageNamePrefix = "openebs-disk-xfs_quota"
```
akhilerm: Does the disk image name always has the xfs_quota in it?

tests/disk/disk.go
```diff
@@ -82,28 +77,30 @@ func (disk *Disk) createDiskImage() error {
...
+	createLoopDeviceCommand := "losetup -P " + "/dev/" + getLoopDevName() + " " + disk.imageName + " --show"
+	var devicePath []byte
+	devicePath, err = RunCommand(createLoopDeviceCommand)
+	if err != nil {
+		return fmt.Errorf("error creating loop device. Error : %v", err)
 	}
+	// Trim trailing newline
+	devicePath = devicePath[:len(devicePath)-1]
+
+	disk.DiskPath = string(devicePath)
```
akhilerm, line 103: Why is this conversion needed here? Its already a string.

tests/disk/disk.go
```diff
@@ -118,59 +115,113 @@ func getLoopDevName() string {
...
+	_, err = RunCommand(wipefsDeviceCommand)
+	if err != nil {
+		return fmt.Errorf("error wiping disk %s. Error: %v",
+			disk.DiskPath, err,
+		)
+	}
+	return nil
+}
+
+func MkdirAll(paths ...string) error {
```
akhilerm: Why cant the `os.MkdirAll()` be used ?

akhilerm (commented):
Given a few more comments @niladrih mostly on some golang specific things.

tests/disk/disk.go
niladrih: It's a byte slice

tests/disk/disk.go
niladrih: Didn't get the question.

akhilerm: The image name is hardcoded with xfs_quota. May be it can be like
```suggestion
	DiskImageNamePrefix = "openebs-disk"
	XFSQuotaDiskImageName = DiskImageNamePrefix + "-" +  "xfs_quota"
```

niladrih: The entire thing is prefixed to a hash generated by tempfile so that the name is unique.

akhilerm: ok. then rename it to `XFSQuotaDiskImageNamePrefix` or something like that, because just the DiskIMageNamePrefix is too generic.

tests/disk/disk.go
niladrih: It can... this is how it was done.... so I went ahead with it.
Moving to os.MkdirAll

niladrih: This was how it was done earlier, so went with it.
Moving to os.MkdirAll()

akhilerm (approved):
LGTM.

PR #148, opened 2022-10-03
chore(doc): Improve documentation for Talos

dbackeus:

Attempting to install the provisioner on Talos without adding privilege to the `openebs` namespace will cause errors since Talos security policy will prevent the provisioner from using `hostPath`

**What this PR does?**:

Adds documentation regarding the need for adding privilege as well as a note of caution that `--preserve` must be used when upgrading Talos to avoid wiping out volumes.

Note: I'm not an expert when it comes to Kubernetes security so feel free to propose a different approach for meeting the Talos security policy. In this PR I provided an example of YAML configuration that is confirmed to work for us using the current version of Talos.

### DIFF
--- a/docs/installation/platforms/talos.md
+++ b/docs/installation/platforms/talos.md
@@ -19,4 +19,19 @@ kubelet:
             - rw
 ```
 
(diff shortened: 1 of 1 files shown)


niladrih:
Thank you for your contribution @dbackeus .

PR #166, opened 2023-12-06
feat(usage): update ua to ga4 analytics

niladrih:
### Refs
- https://github.com/openebs/go-ogle-analytics/pull/5
- https://github.com/openebs/go-ogle-analytics/pull/2

### Description
Moving measurement protocol from UA to GA4.

### API Docs
- https://developers.google.com/analytics/devguides/collection/protocol/ga4/reference
- https://developers.google.com/analytics/devguides/collection/protocol/v1/reference

Fixes:-
https://github.com/openebs/dynamic-nfs-provisioner/issues/175


cmd/provisioner-localpv/app/provisioner.go
```diff
@@ -36,20 +36,29 @@ import (
...
+	pvController "sigs.k8s.io/sig-storage-lib-external-provisioner/v9/controller"
+)
+
+const (
+	// Ping message
+	Ping string = "ping"
+	// DefaultCASType Event application name constant for volume event
+	DefaultCASType string = "localpv"
+	// DefaultUnknownReplicaCount is the default replica count
+	DefaultUnknownReplicaCount string = "replica:<N/A>"
```
Abhinandan-Purkait: replica count should be 1 here since it's localpv

Abhinandan-Purkait (approved):

cmd/provisioner-localpv/app/provisioner.go
niladrih: Good catch! Fixed it. Thanks.

PR #172, opened 2024-02-28
revert!: disable device localpv

Abhinandan-Purkait:
### DEPRECATE LOCAL-PV DEVICE

This PR includes the bits that would contribute to the **_deprecation_** of **_localpv device_**. Henceforth all the dependent charts needed for **_localpv device_** ex. **_NDM_** would be removed. **_Provisioning_** and **_de-provisioning_** of **_local device_** volumes will **_no longer be supported_**.

Since this is a breaking change, we are bumping up the major version from 3.5.x to 4.0.x.


README.md
```diff
@@ -32,33 +32,20 @@ Some of the targetted applications are:
...
-| `v3.3.x` | ✕                  | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               |
-| `v3.4.x` | ✕                  | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               |
-| `v3.5.x` | ✕                  | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               |
-| `HEAD`   | ✕                  | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               | ✓               |
-
+|          | Kubernetes <= 1.19 | Kubernetes 1.20 | Kubernetes 1.21 | Kubernetes 1.22 | Kubernetes 1.23 | Kubernetes 1.24 | Kubernetes 1.25 | Kubernetes 1.26 | Kubernetes 1.27 |
```
niladrih: 1.19 should be the minimum.
We should have a `<= 1.18` table entry with the `✕`
1.19 should have a `✓`

niladrih: Also it should list 3.4, 3.5 and 4.0

niladrih (commented):
Requires changelog directory entry and CHANGELOG.md try.

README.md
Abhinandan-Purkait: We don't want to list 3.4 and 3.5 here right? They have different scopes and features.

niladrih: Fair enough, let's put an entry for HEAD?

niladrih (commented):
Let's remove the `device-integration-test` makefile target, as those would no longer function after the changes to the ginkgo tests.

README.md
avishnu: Is the version change for openebs or localpv? If it's for localpv, I want us to consider releasing a 3.6 or whatever is the next version. Basically, I want us to not tie the openebs umbrella version with the localpv hostpath version.

Abhinandan-Purkait:
> Let's remove the `device-integration-test` makefile target, as those would no longer function after the changes to the ginkgo tests.

Have already removed it. Should I remove the `integration-test` as well? Since we only use `hostpath-integration-test`

tiagolobocastro (approved):
it's not deprecating the code, it's removing it completely right?

`revert!: disable device localpv`

EDIT: maybe disable is more appropriate actually?

README.md
Abhinandan-Purkait: This change is for localpv. Since its a breaking change we decided to bump the major version to 4.x, which happens to have coincided with openebs version. LocalPV versions are anyhow independent of umbrella.

avishnu: Ok, makes sense.

avishnu (approved):
lgtm

niladrih:
> > Let's remove the `device-integration-test` makefile target, as those would no longer function after the changes to the ginkgo tests.
> 
> Have already removed it. Should I remove the `integration-test` as well? Since we only use `hostpath-integration-test`

Yes, it won't work I think.

niladrih (approved):

PR #173, opened 2024-02-28
revert!: remove device localpv from helm charts

Abhinandan-Purkait:
### DEPRECATE LOCAL-PV DEVICE INSTALLATION

This PR includes the bits that would contribute to the **_deprecation_** of **_localpv device_**. Henceforth all the dependent charts needed for **_localpv device_** ex. **_NDM_** would be removed. **_Provisioning_** and **_de-provisioning_** of **_local device_** volumes will **_no longer be supported_**.

Since this is a breaking change, we are bumping up the major version from 3.5.x to 4.0.x.


deploy/helm/charts/README.md
```diff
@@ -96,59 +66,44 @@ helm install openebs-localpv openebs-localpv/localpv-provisioner --namespace ope
...
+| `helperPod.image.registry`                  | Registry for helper image                                                                                                                                                                   | `""`                          |
+| `helperPod.image.repository`                | Image for helper pod                                                                                                                                                                        | `"openebs/linux-utils"`       |
+| `helperPod.image.pullPolicy`                | Pull policy for helper pod                                                                                                                                                                  | `"IfNotPresent"`              |
+| `helperPod.image.tag`                       | Image tag for helper image                                                                                                                                                                  | `3.5.0`                       |
```
niladrih: Let's bump this to 4.0.0 and make the corresponding change in the values as well? Linux-utils follows the localpv versioning scheme.

Abhinandan-Purkait: We would need to release that first maybe then?

Abhinandan-Purkait: Done.

niladrih (approved):

PR #192, opened 2024-07-02
chore(deps): upgrade to ginkgo v2.14.0 and gomega v1.30.0

niladrih:


.github/workflows/pull_request.yml
```diff
@@ -135,12 +121,12 @@ jobs:
...
         run: |
           sudo apt-get update && sudo apt-get install -y xfsprogs quota linux-modules-extra-`uname -r`
-          go install github.com/onsi/ginkgo/ginkgo@v1.16.4
+          go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo@v2.14.0
           sed -e '/openebs-provisioner-hostpath/{N;s/IfNotPresent/Never/}' deploy/kubectl/openebs-operator-lite.yaml | kubectl apply -f -
 
       - name: Integration test
         run: |
-          make hostpath-integration-test
+          make integration-test
```
tiagolobocastro, line 129: what's the difference here?

niladrih: This goes back to v3.5.0 (the version before we released v4.0). [This](https://github.com/openebs/dynamic-localpv-provisioner/blob/v3.5.x/Makefile#L139) is what the integration test targets looked like. There were two separate targets to run integration tests for 'hostpath' mode and 'device' mode separately. And one other mode to run all of the tests.

As a part of the device mode removal PR, the only change made to the tests was that we removed the device mode target and change the `make integration-test` call to `make hostpath-integration-test`.

As a part of this PR, I am removing the device tests and the makefile target for them. This would mean that we run all of the integration tests via the `integration-test` target and they'd all be hostpath tests.

niladrih:
This is such a relief that this upgrade went smoothly!

tiagolobocastro (approved):

PR #194, opened 2024-07-03
feat(chart): add chart changes for v4.1.0

niladrih:


deploy/helm/charts/values.yaml
```diff
@@ -118,7 +120,10 @@ helperPod:
     repository: openebs/linux-utils
     pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
-    tag: 4.0.0
+    tag: 4.1.0
```
tiagolobocastro, line 123: Was there any change to linux-utils?

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

deploy/helm/charts/values.yaml
Abhinandan-Purkait: We created a new tag for the alpine-bash and alpine-sh which also created a new linux-utils but no change in code.

niladrih: We've upgraded the base image for the container to alpine:3.20.1.

Abhinandan-Purkait: Ahh right, the base image is changed. So it's good to have this.

PR #198, opened 2024-07-26
Add BDD spec

niladrih:
This PR adds BDD specs for existing LocalPV Hostpath CI tests.

Also, adds some fixes for quota tests --> adding namespace where the PVC delete operations don't have namespaces set

### DIFF
--- a/tests/bdd/ext4_quota.feature
+++ b/tests/bdd/ext4_quota.feature
@@ -1,0 +1,62 @@
+Feature: Hostpath EXT4 Quota Local PV
+
+  Scenario: HostPath EXT4 Quota Local PV with Unsupported Filesystem
+    Given a sparse file "disk.img"
+    And a loop device is created on top of disk.img
+
+    When a StorageClass with valid EXT4 quota parameters is created
+    Then it should create a StorageClass with the following attributes:
+      | name                | sc-hp-ext4                 |
+      | BasePath            | /path/to/hostpath          |
+      | EXT4QuotaEnabled    | "true"                     |
+      | softLimit           | "20%"                      |
+      | hardLimit           | "50%"                      |
+      | provisionerName     | openebs.io/local           |
+      | volumeBindingMode   | WaitForFirstConsumer       |
+      | reclaimPolicy       | Delete                     |
+
+    When a minix filesystem is written into the loop device
+    And the minix filesystem is mounted with project quota enabled
+    And a PVC "pvc-hp-ext4" is created with the StorageClass "sc-hp-ext4"
+    Then the PVC should be created successfully
+
+    When a Pod is created with PVC "pvc-hp-ext4"
+    Then the Pod should be in pending state
+    And the PVC should be in pending state
+
+    When the Pod "busybox-hostpath" is deleted
+    Then the Pod should be deleted successfully
+
+    When the PVC "pvc-hp-ext4" is deleted
+    Then the PVC should be deleted successfully
+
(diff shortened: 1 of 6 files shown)


tests/bdd/ext4_quota.feature
```diff
@@ -0,0 +1,62 @@
+Feature: Hostpath EXT4 Quota Local PV
+
+  Scenario: HostPath EXT4 Quota Local PV with Unsupported Filesystem
+    Given a sparse file "disk.img"
+    And a loop device is created on top of disk.img
+
+    When a StorageClass with valid EXT4 quota parameters is created
+    Then it should create a StorageClass with the following attributes:
```
tiagolobocastro: who is it?
Are these the parameters specified in the SC? If so, then this needs rewording as it's sounding like these parameters get added by the provisioner somehow

tests/bdd/ext4_quota.feature
```diff
@@ -0,0 +1,62 @@
...
+      | name                | sc-hp-ext4                 |
+      | BasePath            | /path/to/hostpath          |
+      | EXT4QuotaEnabled    | "true"                     |
+      | softLimit           | "20%"                      |
+      | hardLimit           | "50%"                      |
+      | provisionerName     | openebs.io/local           |
+      | volumeBindingMode   | WaitForFirstConsumer       |
+      | reclaimPolicy       | Delete                     |
+
+    When a minix filesystem is written into the loop device
```
tiagolobocastro: :')
I wrote a filesystem based off of minix once

tests/bdd/ext4_quota.feature
```diff
@@ -0,0 +1,62 @@
...
+    And the minix filesystem is mounted with project quota enabled
+    And a PVC "pvc-hp-ext4" is created with the StorageClass "sc-hp-ext4"
+    Then the PVC should be created successfully
```
tiagolobocastro: hmm does this succeed in this case?

tests/bdd/ext4_quota.feature
niladrih: The test creates a StorageClass with these parameters. The Then bit should be in the When, I think?

tests/bdd/ext4_quota.feature
niladrih: I have the minix book :'). And writing a VFS compatible filesystem is also the reason I have it.
Respect +1 for you lord Tiago

tests/bdd/ext4_quota.feature
niladrih: The PVC creation should succeed. The PV creation should fail, also provisioning should fail. Again, this bit should be a part of the `when`. I think all the `then`s belong to what the tested process does.

tests/bdd/ext4_quota.feature
tiagolobocastro: Yeah I think so, like when a storage class is created with the following attributes:
......
.....

tests/bdd/ext4_quota.feature
tiagolobocastro: That's awesome, +1 for you too :D

tests/bdd/ext4_quota.feature
tiagolobocastro: Oh right PV**C** will succeed ofc

tests/bdd/ext4_quota.feature
niladrih: I've used the `Then` directive and its `And`s only when defining codebase behaviour.

Abhinandan-Purkait (approved):

Issue #225, opened 2024-12-24
[bug] pod crashed when creating init pod, new pod always fail because the init pod already exists


styshoo:
**Describe the bug:** A clear and concise description of what the bug is.
dynamic-localpv-provisioner pod crashed(for example: crashed because of renewing lease failed) when creating init pod, new provisioner pod starts, and it continues creating volume. But it always fails because the init pod already exists. So the pvc will always be pending status.

dynamic-localpv-provisioner pods status are as following, one pod restarted.
``` bash
# kubectl get pod -A | grep openebs
kube-system   openebs-hostpath-localpv-provisioner-7f8d6c886-fm6qt       1/1     Running     1             12m
kube-system   openebs-hostpath-localpv-provisioner-7f8d6c886-cjhpk       1/1     Running     0             12m
```

The restarted pod's log is as following, it failed to renew lease, then the pod restarted.
``` log
I1218 02:42:43.386406       1 leaderelection.go:283] failed to renew lease kube-system/openebs.io-local: timed out waiting for the condition
F1218 02:42:43.386436       1 controller.go:889] leaderelection lost
E1218 02:42:43.386611       1 helper_hostpath.go:377] unable to delete the helper pod: client rate limiter Wait returned an error: context canceled
I1218 02:42:46.811347       1 provisioner_hostpath.go:92] Initialize volume pvc-a1f60f52-c3bd-45d6-a708-5a14be2520cc failed: Get "https://172.16.0.1:443/api/v1/namespaces/kube-system/pods/init-pvc-a1f60f52-c3bd-45d6-a708-5a14be2520cc": context canceled
```

The other dynamic-localpv-provisioner pod became the leader, but it failed to continue creating init pod.  
``` log
W1218 02:43:03.897377       1 controller.go:937] Retrying syncing claim "a1f60f52-c3bd-45d6-a708-5a14be2520cc" because failures 0 < threshold 15
[...]

PR #226, opened 2024-12-24
fix: pod crashed when creating init pod, new pod always fail because the init pod already exists

styshoo:
## Pull Request template

dynamic-localpv-provisioner pod crashed(for example: crashed because of renewing lease failed) when creating init pod, new provisioner pod starts, and it continues creating volume. But it always fails because the init pod already exists. So the pvc will always be pending status.

**What this PR does?**:
When creating init pods, it would ignore `pod existes error`.

No.

_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [x] Fixes #<[225](https://github.com/openebs/dynamic-localpv-provisioner/issues/225)>
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests

### DIFF
--- a/cmd/provisioner-localpv/app/helper_hostpath.go
+++ b/cmd/provisioner-localpv/app/helper_hostpath.go
@@ -317,8 +317,8 @@ func (p *Provisioner) launchPod(ctx context.Context, config podConfig) (*corev1.
 		return nil, err
 	}
 	if podList.Items != nil && len(podList.Items) != 0 {
-		klog.Infof("existing helper podList length: %d", len(podList.Items))
-		klog.Infof("existing helper pod: %v", podList.Items[0])
+		klog.V(2).Infof("existing helper podList length: %d", len(podList.Items))
+		klog.V(2).Infof("existing helper pod: %s/%s", podList.Items[0].Namespace, podList.Items[0].Name)
 		return &podList.Items[0], nil
 	}
 
--- a/cmd/provisioner-localpv/app/helper_hostpath.go
+++ b/cmd/provisioner-localpv/app/helper_hostpath.go
@@ -12,6 +12,7 @@ import (
 	hostpath "github.com/openebs/maya/pkg/hostpath/v1alpha1"
 	errors "github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
@@ -215,7 +216,7 @@ func (p *Provisioner) createCleanupPod(ctx context.Context, pOpts *HelperPodOpti
 	config.pOpts.cmdsForPath = append(config.pOpts.cmdsForPath, filepath.Join("/data/", config.volumeDir))
 
 	cPod, err := p.launchPod(ctx, config)
-	if err != nil {
+	if err != nil && !k8serror.IsAlreadyExists(err) {
 		return err
 	}
 
@@ -287,7 +288,7 @@ func (p *Provisioner) createQuotaPod(ctx context.Context, pOpts *HelperPodOption
 	config.pOpts.cmdsForPath = []string{"sh", "-c", fs + checkQuota}
 
 	qPod, err := p.launchPod(ctx, config)
-	if err != nil {
+	if err != nil && !k8serror.IsAlreadyExists(err) {
 		return err
 	}
 
--- a/cmd/provisioner-localpv/app/helper_hostpath.go
+++ b/cmd/provisioner-localpv/app/helper_hostpath.go
@@ -174,12 +174,12 @@ func (p *Provisioner) createInitPod(ctx context.Context, pOpts *HelperPodOptions
 
 	config.pOpts.cmdsForPath = append(config.pOpts.cmdsForPath, filepath.Join("/data/", config.volumeDir))
 
-	_, err := p.launchPod(ctx, config)
+	iPod, err := p.launchPod(ctx, config)
 	if err != nil {
 		return err
 	}
 
-	if err := p.exitPod(ctx, config.podName+"-"+config.pOpts.name); err != nil {
+	if err := p.exitPod(ctx, iPod); err != nil {
 		return err
 	}
 
@@ -215,12 +215,12 @@ func (p *Provisioner) createCleanupPod(ctx context.Context, pOpts *HelperPodOpti
 
 	config.pOpts.cmdsForPath = append(config.pOpts.cmdsForPath, filepath.Join("/data/", config.volumeDir))
 
-	_, err := p.launchPod(ctx, config)
+	cPod, err := p.launchPod(ctx, config)
 	if err != nil {
 		return err
 	}
 
-	if err := p.exitPod(ctx, config.podName+"-"+config.pOpts.name); err != nil {
+	if err := p.exitPod(ctx, cPod); err != nil {
 		return err
 	}
 	return nil
@@ -287,12 +287,12 @@ func (p *Provisioner) createQuotaPod(ctx context.Context, pOpts *HelperPodOption
 		"  rm -rf " + filepath.Join("/data/", config.volumeDir) + " ; exit 1; fi"
 	config.pOpts.cmdsForPath = []string{"sh", "-c", fs + checkQuota}
 
-	_, err := p.launchPod(ctx, config)
+	qPod, err := p.launchPod(ctx, config)
 	if err != nil {
 		return err
 	}
 
-	if err := p.exitPod(ctx, config.podName+"-"+config.pOpts.name); err != nil {
+	if err := p.exitPod(ctx, qPod); err != nil {
 		return err
 	}
 
@@ -371,9 +371,9 @@ func (p *Provisioner) launchPod(ctx context.Context, config podConfig) (*corev1.
 	return hPod, err
 }
 
-func (p *Provisioner) exitPod(ctx context.Context, hPodName string) error {
+func (p *Provisioner) exitPod(ctx context.Context, hPod *corev1.Pod) error {
 	defer func() {
-		e := p.kubeClient.CoreV1().Pods(p.namespace).Delete(ctx, hPodName, metav1.DeleteOptions{})
+		e := p.kubeClient.CoreV1().Pods(p.namespace).Delete(ctx, hPod.Name, metav1.DeleteOptions{})
 		if e != nil {
 			klog.Errorf("unable to delete the helper pod: %v", e)
 		}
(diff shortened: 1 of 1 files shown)


cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -173,12 +174,12 @@ func (p *Provisioner) createInitPod(ctx context.Context, pOpts *HelperPodOptions
 
 	config.pOpts.cmdsForPath = append(config.pOpts.cmdsForPath, filepath.Join("/data/", config.volumeDir))
 
-	iPod, err := p.launchPod(ctx, config)
-	if err != nil {
+	_, err := p.launchPod(ctx, config)
+	if err != nil && !k8serror.IsAlreadyExists(err) {
```
tiagolobocastro, line 178: Do we need to check if the config is compatible with the existing pod?

tiagolobocastro (commented):
Could you please DCO sign the commits?

cmd/provisioner-localpv/app/helper_hostpath.go
styshoo: I'm not sure we need check the config.
But if we do, I have two questions:
1, When do we need to check? Only when this "k8serror.IsAlreadyExists" happens?
2, What kinds of config should be checked?  Compare "pod commands"?

tiagolobocastro: I would probably say we need to compare the entire pod spec? I wonder if the error code here already has such information?
CC @niladrih

styshoo: After some research, I find that it's difficult to compare entire pod spec between the  helperPod and existingPod. Because existingPod would be added lots of default fields by Kubernetes.

For example, the two pods's volume would be as following, the `Type` filed is modified.
```
helperPod.Spec.Volumes[i]: {data {&HostPathVolumeSource{Path:/opt/storage,Type:nil,} nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil}}
existingPod.Spec.Volumes[i]: {data {&HostPathVolumeSource{Path:/opt/storage,Type:*,} nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil nil}}
```

By the way, do we really need to check the compatiblity? Or, do we need to compare the entire pod spec? I didn't find out a good way to check it. Are there any ideas?
@tiagolobocastro

tiagolobocastro: CC @niladrih

niladrih: This could work, I think:
Add labels containing the PVC name and namespace to the init pod and the cleanup pod, e.g.
```
openebs.io/pvc-name: <pvc-name>
openebs.io/pvc-namespace: <pvc-namespace>
```
During `launchPod()`, check to see if there exists a pod with the LabelSelector `openebs.io/pvc-name=<pvc-name>,openebs.io/pvc-namespace=<pvc-namespace>`.

WDYT @styshoo @tiagolobocastro?

styshoo: I'm fine with that.
And I've modified the code, pls help me review it.

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -304,6 +305,20 @@ func (p *Provisioner) launchPod(ctx context.Context, config podConfig) (*corev1.
...
+		"openebs.io/pvc-namespace": p.namespace,
+	}
+	podList, err := p.kubeClient.CoreV1().Pods(p.namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: labels.Set(matchLabels).String(),
+	})
+	if err != nil {
+		return nil, err
+	}
+	if podList.Items != nil && len(podList.Items) != 0 {
+		return &podList.Items[0], nil
```
niladrih: Might be a good idea to `klog.Info` the slice of Pods.

cmd/provisioner-localpv/app/helper_hostpath.go
styshoo: In order to prevent confusion in `init/cleanup`, do we need add an external label? such as :  

``` yaml
openebs.io/helper-pod-name: <pvc-name>
```
or

```
openebs.io/`helper-pod-type` : "init"
```
@tiagolobocastro @niladrih

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/labels"
```
niladrih: This import order seems off. Please run `go fmt cmd/provisioner-localpv/app/helper_hostpath.go`.

cmd/provisioner-localpv/app/helper_hostpath.go
niladrih: Sounds like a good idea to me. How about calling it
```yaml
openebs.io/helper-type: hostpath-init
```
... in case the helper shows up in other openebs projects later..

cmd/provisioner-localpv/app/helper_hostpath.go
styshoo: I have modified the code, please review it for me again. Thanks.

cmd/provisioner-localpv/app/helper_hostpath.go
tiagolobocastro: hmm what should the log say?

styshoo: I've modified it to this:
``` Golang
	if podList.Items != nil && len(podList.Items) != 0 {
		klog.Infof("existing helper podList length: %d", len(podList.Items))
		klog.Infof("existing helper pod: %v", podList.Items[0])
		return &podList.Items[0], nil
	}
```
However, it would print a lot.

tiagolobocastro: Yeah, I don't think we need to do it.

styshoo: So, what should we do next? Remove the log code, or change it from "Info" to "debug"?

tiagolobocastro: @niladrih ?

niladrih: We could put it behind a higher level of verbosity, e.g. `klog.V(2).Info()`. Also maybe all that is required is the name of the pod and the namespace (or maybe not even that). We need not %v print the whole Pod object, I feel.

The reason for this log is to be able to debug contention introduced by more than one init container, by reviewing the logs of the provisioner. Maybe even just printing the number of Pods found at level 2 verbosity is good enough.

styshoo: I've logged the pod count and pod namespace/name, I think `level 2 verbosity` won't print too much.
``` Golang
	if podList.Items != nil && len(podList.Items) != 0 {
		klog.V(2).Infof("existing helper podList length: %d", len(podList.Items))
		klog.V(2).Infof("existing helper pod: %s/%s", podList.Items[0].Namespace, podList.Items[0].Name)
		return &podList.Items[0], nil
	}
```

cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -304,6 +305,23 @@ func (p *Provisioner) launchPod(ctx context.Context, config podConfig) (*corev1.
 	// Helper pods need to create and delete directories on the host.
 	privileged := true
 
+	matchLabels := map[string]string{
+		"openebs.io/pvc-name":      config.pOpts.name,
+		"openebs.io/pvc-namespace": p.namespace,
+		"openebs.io/helper-type":   "hostpath-" + config.podName,
+	}
```
tiagolobocastro: Thinking about this further, this may be problematic for upgrades, if we crash between createPod and exitPod?

Since we haven't changed any of the pod definition on this commit here, I think it's probably fine to just keep as you had initially since we're not making changes to the Pod spec in this PR itself, we can take this up when needed again.
Otherwise I'd suggest to using Apply instead of Create, which should succeed when it already exists and is compatible: <https://github.com/kubernetes/client-go/blob/v0.27.2/dynamic/interface.go#L43> ?

biscout42:
> But it always fails because the init pod already exists

could you help me to understand, is it because `pod` or `pvc` exists? If it is `pvc` , would it make sense to query for `pvc` and in case it is there skip initial pod creation?

niladrih:
>could you help me to understand, is it because pod or pvc exists? If it is pvc , would it make sense to query for pvc and in case it is there skip initial pod creation?

@biscout42 -- Yes, and this is the solution that's currently on this PR.

cmd/provisioner-localpv/app/helper_hostpath.go
niladrih: @styshoo what do you think about the apply solution above?

styshoo: I agree that we can deal with the existing pod without checking.  I viewed  `rancher localpath`, they have same problem, and they just check `k8serror.IsAlreadyExists(err)`, too.

However, I've viewed the Apply interface, `Apply`  interface use `corev1.PodApplyConfiguration` instead of `v1.Pod` in `Create` interface. This interface is much more complicated than Create, and its helper functions (With...) are not as convenient as `Create`. 
The following code will be much more complicated, not as elegant  as  `Create` interface.
``` Golang
	helperPod, err := pod.NewBuilder().
		WithName(config.podName + "-" + config.pOpts.name).
		WithRestartPolicy(corev1.RestartPolicyNever).
```
@niladrih @tiagolobocastro

tiagolobocastro: Alright, let's go with the `k8serror.IsAlreadyExists(err)`. We can look into the `apply` alternative if we start modifying the pod and it becomes a potential issue.

styshoo: OK, I've restored the code to what it was from the beginning.

tiagolobocastro (approved):

niladrih (approved):

styshoo:
@niladrih hi, I noticed  that I haven't do the DCO check, so I just finished it, what should I do next?

niladrih:
@styshoo you could squash the commits into a single one, and add the signature there. I see that the issue is that the DCO check expects a different email address. Do you have the email you've mentioned in the signature, added to your GitHub account?

tiagolobocastro:
> @styshoo you could squash the commits into a single one, and add the signature there. I see that the issue is that the DCO check expects a different email address. Do you have the email you've mentioned in the signature, added to your GitHub account?

DCO is passing? We can squash them ourselves when merging the PR?

PR #238, opened 2025-01-30
feat(charts): add eviction tolerations to the provisioner deployment

niladrih:


deploy/helm/charts/values.yaml
```diff
@@ -48,7 +48,15 @@ localpv:
...
-  tolerations: []
+  tolerations:
+    - effect: NoExecute
+      key: node.kubernetes.io/unreachable
+      operator: Exists
+      tolerationSeconds: 5
+    - effect: NoExecute
+      key: node.kubernetes.io/not-ready
+      operator: Exists
+      tolerationSeconds: 5
```
Abhinandan-Purkait: Hmm, what about addition tolerations? If users adds it via set will it remove these? Can we make it in a way that this toleration always stays not matter whatever extra user adds?

Abhinandan-Purkait (approved):

deploy/helm/charts/values.yaml
Abhinandan-Purkait: Something like this? 

https://github.com/openebs/mayastor-extensions/blob/8ff0e730d216151cf894912b2acdd1db70911a27/chart/templates/_helpers.tpl#L187

niladrih: Right, that would make it easier to upgrade.

niladrih: I've done it a bit differently, somewhat simply. I think this covers all of the cases that the above template function does. What do you think?

deploy/helm/charts/templates/deployment.yaml
```diff
@@ -122,6 +122,14 @@ spec:
 {{- end }}
 {{- if .Values.localpv.tolerations }}
       tolerations:
+        - effect: NoExecute
```
tiagolobocastro: can't we just set the `.Values.localpv.tolerations` value in the chart?
Otherwise this means you can never modify these, and `.Values.localpv.tolerations` become in addition to this

deploy/helm/charts/values.yaml
Abhinandan-Purkait: Looks good to me

deploy/helm/charts/templates/deployment.yaml
Abhinandan-Purkait: Yes that's the intent, we want this to always stay, as it's critical for the pod to move. I mean atleast for mayastor control plane components, but not sure if we want to hold that true for localpv controller.

tiagolobocastro: But what if someone doesn't want them? Mayastor is just a use-case, we can make any requirements there without forcing it on every dynamic user
What if someone wants different seconds for example?

niladrih: I'll move it back to `.Values.localpv.tolerations`.

niladrih: Done. PTAL folks.

tiagolobocastro: I'm going to make a different suggestion.

On mayastor we have a separate field `earlyEvictionTolerations` which can be separately enabled/disabled with the "general" `tolerations`.
Would this achieve on we wanted?
1. It's always enabled by default, even if users sets other `tolerations`
2. Can be disabled if user doesn't want them by clearing `earlyEvictionTolerations`

What do you think?

Abhinandan-Purkait: Sounds good.

niladrih: Alright, yeah that sounds good in this context.

niladrih: Done.

deploy/helm/charts/templates/_helpers.tpl
```diff
@@ -75,3 +75,15 @@ Create the name of the service account to use
...
 {{- end -}}
 {{- end -}}
+
+{{/*
+Creates the tolerations based on the global tolerations, with early eviction
+Usage:
+{{ include "tolerations_with_early_eviction" . }}
+*/}}
+{{- define "tolerations_with_early_eviction" -}}
+{{- toYaml .Values.earlyEvictionTolerations | nindent 8 }}
```
tiagolobocastro: Doesn't this need an `if` as well?

tiagolobocastro (approved):

deploy/helm/charts/templates/_helpers.tpl
niladrih: Sure, that's a good idea.

PR #260, opened 2025-04-29
feat(helm chart): refactor labels and add logging labels

Abhinandan-Purkait:
- Adds label selectors for logging.
- Removes redundant labels.

### DIFF
--- a/deploy/helm/charts/values.yaml
+++ b/deploy/helm/charts/values.yaml
@@ -136,6 +136,9 @@ helperPod:
 # Additional labels to add to all chart resources
 extraLabels: {}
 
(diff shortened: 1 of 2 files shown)


deploy/helm/charts/values.yaml
```diff
@@ -13,6 +13,9 @@ rbac:
   # rbac.pspEnabled: `true` if PodSecurityPolicy resources should be created
   pspEnabled: false
 
+loggingLabels:
```
niladrih: Let's move this next to the `extraLabels` key?

niladrih (approved):

PR #271, opened 2025-06-02
build: add support for updating the image tag on the helm chart README.md

niladrih:
This script doesn't update the README.md image tag. This change adds an sed command to do that.

### DIFF
--- a/scripts/update-chart-version.sh
+++ b/scripts/update-chart-version.sh
@@ -138,6 +138,14 @@ update_chart_yaml() {
 
   yq_ibl ".version = \"$VERSION\" | .appVersion = \"$APP_VERSION\"" "$CHART_YAML"
   yq_ibl ".localpv.image.tag = \"$VERSION\"" "$VALUES_YAML"
+
+  if sed --version >/dev/null 2>&1; then
+    # GNU sed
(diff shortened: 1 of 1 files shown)


pchandra19 (approved):
sed subsitution lgtm. Please take a look at shell checks and see if you need update.

scripts/update-chart-version.sh
```diff
@@ -138,6 +138,14 @@ update_chart_yaml() {
 
   yq_ibl ".version = \"$VERSION\" | .appVersion = \"$APP_VERSION\"" "$CHART_YAML"
   yq_ibl ".localpv.image.tag = \"$VERSION\"" "$VALUES_YAML"
+
+  if sed --version >/dev/null 2>&1; then
+    # GNU sed
+    sed -i "s/| \`localpv\.image\.tag\`[[:space:]]*| Image tag for LocalPV Provisioner[[:space:]]*| \`[^\\\`]*\`[[:space:]]*|/| \`localpv.image.tag\`                         | Image tag for LocalPV Provisioner                                                                                                                                                           | \`$APP_VERSION\`                       |/" "$README_MD"
```
tiagolobocastro, line 144: Shouldn't helm-docs do this?

niladrih: Yes, but we don't have helm-docs here, and we'd need to comment all the fields appropriately. Also we'd have to add the git hook. I was just being lazy for this one off thing. Do you recommend the helm docs way? I could do the whole thing.

tiagolobocastro (approved):

scripts/update-chart-version.sh
tiagolobocastro: For now it's fine, but ideally would be good to have this helm-doc driven yes

PR #284, opened 2025-09-14
Update deployment.yaml

aminmaghsodi:
We need deployments have their specific namespaces.

 if you run commands:
helm show values openebs/openebs > values.yaml
helm template openebs openebs/openebs --namespace openebs -f values.yaml > openebs-generated-values.yaml

You see many blocks (for example this Deployment block) does not have namespace: openebs

### DIFF
--- a/deploy/helm/charts/templates/deployment.yaml
+++ b/deploy/helm/charts/templates/deployment.yaml
@@ -4,6 +4,7 @@ kind: Deployment
 metadata:
   name: {{ template "localpv.fullname" . }}
   {{- with .Values.localpv.annotations }}
+  namespace: {{ .Release.Namespace }}
   annotations: {{ toYaml . | nindent 4 }}
   {{- end }}
   labels:


deploy/helm/charts/templates/deployment.yaml
```diff
@@ -4,6 +4,7 @@ kind: Deployment
 metadata:
   name: {{ template "localpv.fullname" . }}
   {{- with .Values.localpv.annotations }}
+  namespace: {{ .Release.Namespace }}
```
tiagolobocastro, line 7: Isn't this just default? From your description thought you'd want to pass some other var for the namespace

niladrih:
@aminmaghsodi if you execute your helm install command with `-n <your-namespace` it'd have the same effect. Why do it this way?

aminmaghsodi:
@niladrih 
if you run commands:
helm show values openebs/openebs > values.yaml
helm template openebs openebs/openebs --namespace openebs -f values.yaml > openebs-generated-values.yaml

You see many blocks (e.g. Deployment block) does not have namespace: openebs

deploy/helm/charts/templates/deployment.yaml
tiagolobocastro: Got it now, for the helm template... could you please also add it to the service account?

tiagolobocastro: @aminmaghsodi a gentle ping :)

PR #310, opened 2026-01-27
refactor(quota): remove use of findmnt and remove unnecessary mounts

niladrih:
findmnt discovers the actual mountpath for a given path. This is helpful if the
path is a bind mount. However, this doesn't concern us as filesystem quota options
work with bind mounts just fine. There is no need to use findmnt in this case.

This change also fixes a bug where volumes could only be created on the path specified
in the .Values.localpv.basepath. StorageClass BasePath doesn't change this configuration,
This change mounts the host / path on the Node Deployment mode, so that any path specified
on the StorageClass may work.


cmd/provisioner-localpv/app/helper_hostpath.go
```diff
@@ -213,28 +217,59 @@ func (p *Provisioner) createCleanupPod(ctx context.Context, pOpts *HelperPodOpti
...
+
+	// Host paths (used for findmnt which reads host's mountinfo)
+	// findmnt needs the actual host path, not the /host-prefixed container path
+	hostParentDir := config.parentDir
+	hostVolumePath := filepath.Join(config.parentDir, config.volumeDir)
+
+	// Build cleanup script using findmnt strategy (same as local_volume_manager.go)
+	// This properly handles bind mounts by finding the actual filesystem mount point
+	// Note: findmnt uses host paths, but we access files via /host prefix
+	cleanupScript := fmt.Sprintf(`
```
tiagolobocastro: Can we reuse the same code as the node deployment now?

niladrih: Hmm, yes!

tiagolobocastro (approved):
since we don't have any tests for the quota, please test manually that it works, including the cleanup

cmd/provisioner-localpv/app/local_volume_manager.go
```diff
@@ -36,6 +36,12 @@ type LocalVolumeManager struct {
 	mu *sync.Mutex
 }
 
+const (
+	// HostPathPrefix is the mount point where the host root filesystem is mounted
+	// in the node DaemonSet. This allows the provisioner to access any path on the host.
+	HostPathPrefix = "/host"
```
tiagolobocastro, line 42: Can you also add a comment on the helm chart enable for this, this should be very visible...
Perhaps a future improvement could be to modify the mounts depending on the storage classes

tiagolobocastro (approved):

cmd/provisioner-localpv/app/local_volume_manager.go
niladrih: StorageClasses could demand new paths dynamically. We couldn't prepare for this at install time when starting the Node deployments.

I'll add the comment.

PR #312, opened 2026-02-02
Update sigs.k8s.io/sig-storage-lib-external-provisioner & CI refactor

### LINKED ISSUE #279: Warnings about accessing Endpoints instead of EndpointSlices on Kubernetes v1.33+
My logs for the `localpv-provisioner` pod has an endless stream of warnings about fetching `Endpoints` instead of `EndpointSlice` on my v1.33 k8s cluster:

https://kubernetes.io/blog/2025/04/24/endpoints-deprecation/

```
W0613 12:12:44.075828       1 warnings.go:70] v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1
 EndpointSlice
```

Is there a mechanism to at least silence these warnings as they fill up the disk and add some cpu overhead to constantly generate?

De-duplicating these warnings would also be preferred: https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling

### LINKED ISSUE #287: Provisioner's SupportsBlock() function asserts true when it shouldn’t
**Describe the bug:** The SupportsBlock() function should return a 'false' as the support for block volumes was dropped with the removal of the 'device' mode provisioner.

https://github.com/openebs/dynamic-localpv-provisioner/blob/ed848371268ea3d22a9be9e344a703ae006e4ffd/cmd/provisioner-localpv/app/provisioner.go#L75-L77

This provisioner runs mkdir and other similar commands against filesystems. It definitely does not support block volumes.

**Environment details:**
- OpenEBS version (use `kubectl get po -n openebs --show-labels`): 4.3.3
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`): Ubuntu 22.04.4 LTS
- kernel (e.g: `uname -a`): 5.15.0-140-generic
- others:

### LINKED ISSUE #3925: v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice
## Description

Kubernetes v1.33+ deprecates `core/v1 Endpoints` in favor of `discovery.k8s.io/v1 EndpointSlice`. OpenEBS components currently reference or manage `Endpoints` directly, which will eventually break compatibility with future Kubernetes versions. Migrating to use `EndpointSlice` ensures forward compatibility and better scalability.

## Context

We are upgrading to Kubernetes v1.33+ in our environments, and deprecation warnings are triggered due to OpenEBS usage of `v1 Endpoints`. This change is important to maintain cluster stability and future support. Transitioning to `EndpointSlice` will also benefit the broader OpenEBS user community by ensuring compliance with the latest Kubernetes APIs and improving networking performance at scale.

## Possible Solution

* Audit all OpenEBS components (e.g., Jiva, cStor, NDM) for usage of `core/v1 Endpoints`.
* Refactor controllers or Helm charts to reference or create `discovery.k8s.io/v1 EndpointSlice` instead.
* Add feature detection for clusters < v1.33 if backward compatibility is needed.

## Screenshots

N/A

## Additional Evidence

In our Kubernetes 1.33+ cluster, OpenEBS creates a deprecated `core/v1 Endpoints` resource:

```bash
$ kubectl get endpoints -A | grep openebs
openebs-system      openebs.io-local

$ kubectl get endpoints -A -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.namespace}{"\t"}{.metadata.ownerReferences[*].kind}{"\n"}{end}' | grep -i openebs
openebs.io-local        openebs-system
```

The local PV provisioner appears to be the source:

```bash
$ kubectl get deploy -n openebs-system
openebs-localpv-provisioner   quay.io/openebs/provisioner-localpv:4.2.0
```

Observed Kubernetes warning:

```plaintext
Warning: v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice
```

niladrih:
Fixes #279
Fixes https://github.com/openebs/dynamic-localpv-provisioner/issues/287
Fixes https://github.com/openebs/openebs/issues/3925


.github/workflows/build_and_push.yml
```diff
@@ -22,6 +22,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up Go 1.24.12
+        uses: actions/setup-go@v6
+        with:
+          go-version: 1.24.12
+          cache: false
```
tiagolobocastro: when to use false or true? We have a mix in the workflows

shell.nix
tiagolobocastro, line 1: we're removing go? But then this shell doesn't work?

.github/workflows/build_and_push.yml
```diff
@@ -22,6 +22,12 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up Go 1.24.12
```
tiagolobocastro: Would it be possible to keep all these versions in line with lvm/zfs ?

.github/workflows/build_and_push.yml
niladrih: It's to decide if it should download the dependencies specified in the go.mod file every time it's being built/executed.

Seems to be a good idea to leave it to the default `true` unless it gives us problems.

shell.nix
niladrih: The shell works. Go is set up in the environment outside the shell. The shell doesn't set it up. Doing this keeps the go setup consistent across all workflows. This one specific workflow was setting up go via nix, while the rest were using actions/setup-go.

.github/workflows/build_and_push.yml
niladrih: This requires 1.24, if we want to upgrade to sig.k8s.io/sig-storage-lib-external-provisioner/v13. Went with 1.24.12, because that's the latest path for 1.24.

The zfs-localpv and lvm-localpv repos are at 1.24 as well. At 0 patch. I'll raise the PRs to get them to use a newer patch.

shell.nix
niladrih: On second thought, the nix shell should be self-sufficient. I've put the go back into the shell. Removed the actions/setup-go on the specific workflows which use this shell.

.github/workflows/build_and_push.yml
niladrih: https://github.com/openebs/lib-csi/pull/30
https://github.com/openebs/zfs-localpv/pull/694
https://github.com/openebs/lvm-localpv/pull/445

.github/workflows/build_and_push.yml
tiagolobocastro: Doesn't seem to work?
>  Restore cache failed: Dependencies file is not found in /home/runner/work/dynamic-localpv-provisioner/dynamic-localpv-provisioner. Supported file pattern: go.sum

.github/workflows/build_and_push.yml
```diff
@@ -22,28 +22,26 @@ jobs:
...
+          nix-shell --run "ct lint --config ct.yaml"
 
   unit-test:
     runs-on: ubuntu-latest
     needs: ['lint']
     steps:
-      - name: Set up Go 1.19
-        uses: actions/setup-go@v5
+      - name: Set up Go 1.24.12
+        uses: actions/setup-go@v6
```
tiagolobocastro: You could also mirror the zfs/lvm changes here fully, and also add actions for this, or can be taken after

shell.nix
```diff
@@ -1,28 +1,31 @@
...
-    which
-    curl
-    cacert
-    crane
-    util-linux
-    jq
-    nixos-shell
-  ] ++ pkgs.lib.optional (builtins.getEnv "IN_NIX_SHELL" == "pure") [ docker-client ];
+  buildInputs =
+    with pkgs;
```
tiagolobocastro, line 9: is this formatted with nixpkgs-fmt?

tiagolobocastro (approved):

shell.nix
niladrih: Formatted with nixfmt instead.

.github/workflows/build_and_push.yml
niladrih: Refactored the CI to use reusable workflows.

.github/workflows/build_and_push.yml
niladrih: `cache: false` seems to solve this. Using `cache: false`.

PR #321, opened 2026-05-10
feat: add configurable worker threads and helper pod timeout

kotyara85:
## Pull Request template

Make controller configurable

**What this PR does?**:
Add two new environment variables to tune provisioning performance:

- OPENEBS_IO_WORKER_THREADS: number of concurrent worker goroutines for processing PVC create and PV delete events (default: 4). Higher values increase throughput when many PVCs are created simultaneously.

- OPENEBS_IO_HELPER_POD_TIMEOUT: maximum seconds to wait for a helper pod (init, cleanup, quota) to complete before timing out (default: 120). Lower values reduce wasted time when helper pods cannot be scheduled.

Previously both values were hard-coded with no way to configure them.

No

Yes

### DIFF
--- a/cmd/provisioner-localpv/app/env.go
+++ b/cmd/provisioner-localpv/app/env.go
@@ -44,6 +44,17 @@ const (
 	//
 	// This environment variable is set via kubernetes downward API.
 	OpenebsServiceAccount string = "OPENEBS_SERVICE_ACCOUNT"
+
+	// ProvisionerWorkerThreads is the environment variable that controls the
+	// number of concurrent worker goroutines for processing PVC create and
+	// PV delete events. Higher values increase provisioning throughput when
+	// many PVCs are created simultaneously. Default is 4.
+	ProvisionerWorkerThreads string = "OPENEBS_IO_WORKER_THREADS"
+
+	// ProvisionerHelperPodTimeout is the environment variable that controls
+	// the maximum number of seconds to wait for a helper pod (init, cleanup,
+	// quota) to complete before timing out. Default is 120.
+	ProvisionerHelperPodTimeout string = "OPENEBS_IO_HELPER_POD_TIMEOUT"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


tiagolobocastro (approved):

cmd/provisioner-localpv/app/env.go
```diff
@@ -44,6 +44,17 @@ const (
...
+	// ProvisionerWorkerThreads is the environment variable that controls the
+	// number of concurrent worker goroutines for processing PVC create and
+	// PV delete events. Higher values increase provisioning throughput when
+	// many PVCs are created simultaneously. Default is 4.
+	ProvisionerWorkerThreads string = "OPENEBS_IO_WORKER_THREADS"
+
+	// ProvisionerHelperPodTimeout is the environment variable that controls
+	// the maximum number of seconds to wait for a helper pod (init, cleanup,
+	// quota) to complete before timing out. Default is 120.
+	ProvisionerHelperPodTimeout string = "OPENEBS_IO_HELPER_POD_TIMEOUT"
```
niladrih, line 57:
```suggestion
	ProvisionerHelperPodTimeout string = "OPENEBS_IO_HELPER_POD_TIMEOUT_SECS"
```

cmd/provisioner-localpv/app/env.go
```diff
@@ -77,3 +88,19 @@ func getOpenEBSImagePullSecrets() string {
 func getNodeName() string {
 	return menv.Get("NODE_NAME")
 }
+
+func getWorkerThreads() int {
+	val, err := k8sEnv.GetInt(ProvisionerWorkerThreads, 4)
+	if err != nil || val < 1 {
```
niladrih, line 94: I don't think we should work with less than 2 workers.
```suggestion
	if err != nil || val < 2 {
```

cmd/provisioner-localpv/app/env.go
```diff
@@ -77,3 +88,19 @@ func getOpenEBSImagePullSecrets() string {
...
+		return 4
```
niladrih, line 95: nit: we should log that we're falling back to a default value, as the input is not acceptable.

cmd/provisioner-localpv/app/env.go
```diff
@@ -77,3 +88,19 @@ func getOpenEBSImagePullSecrets() string {
...
+	}
+	return val
+}
+
+func getHelperPodTimeout() int {
+	val, err := k8sEnv.GetInt(ProvisionerHelperPodTimeout, 120)
+	if err != nil || val < 1 {
+		return 120
```
niladrih, line 103: nit: we should log that we're falling back to a default value, as the input is not acceptable.

niladrih (commented):
Also, please rebase your branch. Let me know if you'd need help rebasing. GitHub UI branch update can seem like a rebase, but it's a git pull like update.

niladrih (approved):
Merging this.

PR #322, opened 2026-05-11
refactor(quota): move quota scripts to its own file

niladrih:
The quota setup and cleanup scripts are embedded in code and this makes it less testable and readable. This change moves it to its own file and embeds the script into the code at compile time.

## Pull Request template

The quota setup and cleanup scripts are embedded in code and this makes it less testable and readable. This change moves it to its own file and embeds the script into the code at compile time.

**What this PR does?**:
Moves the embedded shell scripts to its own file. The two apply and cleanup functions are triggered via different arguments.

It's a refactor, shouldn't required upgrade changes.

I'll update after I do more manual testing. Opening the PR for review in the mean time.

_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Commit has unit tests

### DIFF
--- a/cmd/provisioner-localpv/app/quota.sh
+++ b/cmd/provisioner-localpv/app/quota.sh
@@ -1,0 +1,95 @@
+#!/bin/sh
+# OpenEBS dynamic-localpv-provisioner quota helper.
+#
+# Applies or removes a filesystem project quota on a directory whose parent
+# resides on an XFS or EXT2/3/4 filesystem with project quotas enabled.
+#
+# Usage:
+#   quota.sh apply   --parent <path> --volume <path> --soft-kb <n> --hard-kb <n>
+#   quota.sh cleanup --parent <path> --volume <path>
+
+set -e
+
+die() { echo "$@" >&2; exit 1; }
+
+usage() {
+    cat >&2 <<'EOF'
+Usage:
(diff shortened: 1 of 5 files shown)


cmd/provisioner-localpv/app/quota.sh
tiagolobocastro, line 1: Just a thought, could we do this in go directly?

niladrih: For nodeDeployment mode -- sure.

The helperPod mode runs the openebs/linux-utils container on the hosts where the volume is created/deleted. We'd have to deploy the provisioner-localpv container to run in those cases instead of linux-utils. The proivisoner-localpv container is ~47 MiB, the linux-utils container is ~24MiB.

Or, would it be better if we create a new lighter binary to do this instead?

tiagolobocastro: We could have a lighter one, but even the localpv one is probably fine?

niladrih: I'm also fine with having the provisioner-localpv do it. I'll make the change.

PR #323, opened 2026-05-12
fix(helper-pod): serialize quota project-ID allocation via flock

niladrih:
Two helper pods racing on the same node can read the same max projid from repquota/xfs_quota, increment to the same value, and bind two volumes to one quota bucket.

NodeDeployment mode serializes this with a Go mutex in LocalVolumeManager, but HelperPod mode runs each operation in a separate pod, so the lock has to live on the host. Wrap the generated apply and cleanup scripts in a subshell that holds an exclusive flock on <parentDir>/.openebs-quota.lock, gated by a new UseHostLock field on QuotaScriptConfig. The lock fd is scoped to the subshell, so it cannot collide with anything in the caller's environment, and the kernel releases the lock when the subshell exits.

### DIFF
--- a/cmd/provisioner-localpv/app/quota_scripts.go
+++ b/cmd/provisioner-localpv/app/quota_scripts.go
@@ -22,91 +22,119 @@ type QuotaScriptConfig struct {
 	// For HelperPod: use "" (parentDir is already mounted at /data)
 	// For DaemonSet: use "/host" (host root is mounted at /host)
 	HostPathPrefix string
+	// UseHostLock serializes the script via flock on a lockfile in the parent
+	// directory. Set for HelperPod mode; NodeDeployment relies on an in-process
+	// mutex instead.
+	UseHostLock bool
 }
 
-// GenerateQuotaApplyScript generates a shell script to apply filesystem quota
-// This script:
-// 1. Detects filesystem type (XFS or EXT4)
-// 2. Applies project quota with the specified limits
-func GenerateQuotaApplyScript(cfg QuotaScriptConfig) string {
-	// Extract numeric values for EXT4 setquota (it expects plain numbers in KB blocks, not with suffix)
-	extSoftLimit := strings.TrimSuffix(cfg.SoftLimitGrace, "k")
-	extHardLimit := strings.TrimSuffix(cfg.HardLimitGrace, "k")
-
-	// Build paths with optional prefix
-	parentPath := cfg.ParentDir
-	volumePath := filepath.Join(cfg.ParentDir, cfg.VolumeDir)
+// quotaPaths resolves the parent and volume paths from cfg, applying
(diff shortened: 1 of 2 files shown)


cmd/provisioner-localpv/app/quota_scripts.go
```diff
@@ -22,91 +22,119 @@ type QuotaScriptConfig struct {
...
 	}
+	return `# Serialize quota operations on this host.
+LOCKFILE="$PARENT_PATH/.openebs-quota.lock"
+LOCK_WAIT_SECONDS=60
+if ! command -v flock >/dev/null 2>&1; then
+    echo "flock(1) not found in PATH; cannot serialize quota operations on $LOCKFILE" >&2
+    exit 1
+fi
+# Pre-flight the lockfile path: exec 9>FILE aborts the shell on failure
+# (busybox ash does not honour ||), so probe with touch first.
```
tiagolobocastro, line 58: > busybox ash does not honour ||

What does this mean?

the touch as the first op seems similar to what we have [here](https://github.com/openebs/rawfile-localpv/blob/develop/rawfile/utils/lock.py).
what is `--` doing?

niladrih: > what is -- doing?

It's trivial. It's marking end of CLI options, and start of arguments.

> the touch as the first op seems similar to what we have [here](https://github.com/openebs/rawfile-localpv/blob/develop/rawfile/utils/lock.py).

Yes, it's trying to fail fast in case the file is not accessible for some reason.

> > busybox ash does not honour ||
>
> What does this mean?

In an ash shell, which is following strict POSIX, the script aborts immediately when the exec command fails. Apparantly the exec command is classified as a special built-in command, and the posix rules for this is to exit immediately. Bash is not strictly POSIX compliant, this is well known, and there the command returns a non-zero exit code that is consumed by the OR and then we see the second operand... etc.. etc.
You may try this to verify this behaviour
`exec 9>/no/such/dir/x || echo CAUGHT; echo END`

And I suspect one might be inclined to ask 'why is this required? are we on an ash shell?'. So, no it's not required. It may be removed and nothing may change here because of it. Recently, in a conversation with a work acquaintance, we switched out an ubuntu base image for an alpine one, with sufficient testing of course. The thing is, the visibility of these nuances in these standard base images is narrow. Not comprehensively known. And I'd like to think, it could have been busybox, and no one would know, and the tests won't fail, because there are no tests. I know I may be ill informed and busybox is no longer a usable base image in the current day, but it used to be, and I don't know if someday some handy busybox tool may pursuade someone to switch out the alpine on here for buxsybox, ash, whatever. 3 lines of checking didn't feel like a whole lot. So... I can remove it, not required, didn't think too much of it.

tiagolobocastro (approved):
Fine but in general I'd want to move away from this sell stuff, very hard to read and make changes safely.

PR #328, opened 2026-05-18
fix(analytics): emit events once per release, not per pod

niladrih:
Previously, helperPod mode re-emitted `install` on every pod restart and
nodeDeployment mode emitted install/ping/heartbeat from every node,
inflating event counts N-fold per cluster.

Both modes now gate `install` on a Helm-managed ConfigMap
(<fullname>-analytics-state, via OPENEBS_IO_ANALYTICS_STATE_CM) that
helm uninstall removes, so reinstalls correctly re-fire install while
pod restarts and leader transitions do not.

nodeDeployment mode runs install/ping/heartbeat under client-go leader
election against a dedicated Lease (<fullname>-analytics, via
OPENEBS_IO_ANALYTICS_LEASE) so only one pod cluster-wide emits.


Abhinandan-Purkait:
@copilot Can configmap be avoided here?

Abhinandan-Purkait:
@copilot Revert the changes about removing the configmap, it was a question not an action item.

cmd/provisioner-localpv/app/start.go
```diff
@@ -160,3 +183,149 @@ func isLeaderElectionEnabled(ctx context.Context, nodeDeployment bool) bool {
...
+	go runPingLoop(ctx, Heartbeat, true)
+}
+
+// runAnalyticsHelperPod runs the analytics path for helper-pod (Deployment) mode.
+// Single pod, no leader election; install is still CM-gated so a pod restart
+// does not re-emit it.
+func runAnalyticsHelperPod(ctx context.Context, kubeClient kubernetes.Interface, namespace string) {
+	defer recoverAnalytics("helperPod")
+	startAnalyticsEmitters(ctx, kubeClient, namespace)
+}
```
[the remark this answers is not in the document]
niladrih: This is not required for the time being.

cmd/provisioner-localpv/app/start.go
```diff
@@ -160,3 +183,149 @@ func isLeaderElectionEnabled(ctx context.Context, nodeDeployment bool) bool {
...
+	if err != nil {
+		return errors.Wrap(err, "marshal install-at patch")
+	}
+	if _, err := kubeClient.CoreV1().ConfigMaps(namespace).Patch(
+		ctx, cmName, types.MergePatchType, patch, metav1.PatchOptions{},
+	); err != nil {
+		return errors.Wrap(err, "patch analytics state configmap")
+	}
+	klog.V(2).Infof("analytics: install emitted; marked %q in %q", AnalyticsInstalledAtKey, cmName)
+	return nil
```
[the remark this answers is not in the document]
niladrih: The once-per-release need not be a strict guarantee, it's fine.

Abhinandan-Purkait (approved):

cmd/provisioner-localpv/app/start.go
```diff
@@ -160,3 +183,149 @@ func isLeaderElectionEnabled(ctx context.Context, nodeDeployment bool) bool {
...
+	if err != nil {
+		klog.Errorf("analytics: leader elector: %v", err)
+		return
+	}
+	le.Run(ctx)
+}
+
+// maybeEmitInstall sends the install event once per Helm release, gated by a
+// ConfigMap whose name is provided via OPENEBS_IO_ANALYTICS_STATE_CM. The chart
+// creates the CM empty on helm install and removes it on helm uninstall, so a
```
tiagolobocastro, line 264: This means it's actually tracking install+reinstall which is different than rawfile and also different than existing ones, so we may get different numbers
IIRC we wanted a separate reinstall counter

cmd/provisioner-localpv/app/start.go
```diff
@@ -160,3 +183,149 @@ func isLeaderElectionEnabled(ctx context.Context, nodeDeployment bool) bool {
...
+	if _, done := cm.Data[AnalyticsInstalledAtKey]; done {
+		return nil
+	}
+
+	// Send before marking: losing an install on Send failure is better than a
+	// permanent loss on Patch failure after a successful Send.
+	emitInstall()
+
+	patch, err := json.Marshal(map[string]any{
+		"data": map[string]string{AnalyticsInstalledAtKey: time.Now().UTC().Format(time.RFC3339)},
```
tiagolobocastro, line 301: are the keys named the same as the rawfile? Especially if we want to port this other engines I'd like to see consistency here, otherwise it's the same loop on repeat, locals implementing things all differently and no one knows why

cmd/provisioner-localpv/app/start.go
```diff
@@ -160,3 +183,149 @@ func isLeaderElectionEnabled(ctx context.Context, nodeDeployment bool) bool {
...
+	}
+
+	cm, err := kubeClient.CoreV1().ConfigMaps(namespace).Get(ctx, cmName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		// Env var set but the CM is absent from the cluster — chart
+		// updated without the analytics-state template, or someone
+		// deleted it. Same fallback as no-env-var: emit without state
+		// tracking.
+		klog.V(2).Infof("analytics: state ConfigMap %q not found; emitting install without state tracking", cmName)
+		emitInstall()
```
tiagolobocastro, line 286: Would it be better to not send it, since this should be an unexpected state?

tiagolobocastro (commented):
What happens on pod restart, does it re-acquire the lease straight away?

cmd/provisioner-localpv/app/start.go
niladrih: There's no harm in doing this?

cmd/provisioner-localpv/app/start.go
niladrih: The only key is `installed-at`

cmd/provisioner-localpv/app/start.go
niladrih: That's right. I am going to raise a PR to make the behaviour same as rawfile.

cmd/provisioner-localpv/app/start.go
tiagolobocastro: On rawfile it's `install-ts` I think

cmd/provisioner-localpv/app/start.go
tiagolobocastro: Other than again repeating the install event if it gets deleted?

tiagolobocastro:
Also what happens with the ping when a leader is loses (ping cancelled), and then becomes leader again, does the ping restart from 0 again?
Rawfile uses `last-ping-ts` to check when it has to send it again IIRC

PR #335, opened 2026-05-29
feat: globaloverride for gaid and gakey

krishnaGajabi:
Why is this PR required? What issue does it fix?:
This PR implements Global Override behvaiour for gaid and gakey to be consistent with other local engine charts

What this PR does?:
This PR implements Global Override behvaiour for gaid and gakey to be consistent with other local engine charts

Does this PR require any upgrade changes?:
No

If the changes in this PR are manually verified, list down the scenarios covered::
did a helm template test on the chart manually

without anything set: no values are rendered
with global and local values set: global values win
with local values set: local is applied
Repeated the above procedure with `localpv.nodeDeployment.enabled=true` and observed similar results

### DIFF
--- a/deploy/helm/charts/values.yaml
+++ b/deploy/helm/charts/values.yaml
@@ -2,6 +2,15 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# NOTE: This option is required so that .global.imageRegistry works as an override
+#       for the local image registry values. By default, this option is set to false
+#       and that's to retain backwards compatibility with v4.4.0 where .global.imageRegistry
+#       was introduced as the default and the local options were overrides for the global default.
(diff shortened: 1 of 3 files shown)


niladrih (approved):

deploy/helm/charts/values.yaml
```diff
@@ -14,9 +23,6 @@ global:
     # Global override for analytics
```
tiagolobocastro, line 23: Should we add the empty `gaId` and `gaKey` with a comment here as well as the README?

krishnaGajabi: I think users are not going to use these values a lot, so we better not mention them?

tiagolobocastro: Fair enough then

tiagolobocastro (approved):

PR #342, opened 2026-06-17
feat(client): make Kubernetes API client QPS and burst configurable

oneubauer:

Follow-up to #321 (configurable worker threads). #321 made the provisioner's
worker count tunable via `OPENEBS_IO_WORKER_THREADS`, but all workers share a
single Kubernetes clientset, and that clientset's rate limiter is still pinned
to the client-go defaults (QPS 5 / burst 10). Raising the worker count alone
therefore just creates more goroutines that contend for the same client-side
throttle — under bursty PVC create/delete load the provisioner self-throttles
before the API server's own API Priority & Fairness ever comes into play.

This PR exposes the client's QPS and burst so operators running at scale can
raise them alongside `workers`.

**What this PR does?**:

Adds two new, optional environment variables (and matching Helm values) to tune
the rate limiter on the shared Kubernetes API client:

- `OPENEBS_IO_CLIENT_QPS`: maximum sustained queries/sec to the API server.
- `OPENEBS_IO_CLIENT_BURST`: maximum burst above QPS.

Wiring:
- `cmd/provisioner-localpv/app/env.go`: env-var names + `getClientQPS()` /
  `getClientBurst()` getters (functional `menv` lookup, same pattern as the
  existing worker-thread/timeout getters).
- `pkg/kubernetes/client/client.go`: `WithQPS()` / `WithBurst()` functional
  options applied to the `rest.Config` before the clientset is built.
- `cmd/provisioner-localpv/app/start.go`: reads the env vars and passes the
  options through when constructing the client.
- Helm chart: `localpv.controller.clientQPS` / `localpv.controller.clientBurst`
  values, conditionally injected into the Deployment and DaemonSet env, plus
  README parameter docs.

**Default behavior is unchanged.** When the variables are unset (or empty), no
override is applied and client-go's own defaults (QPS 5 / burst 10) remain in
effect. The env vars are only injected by the chart when explicitly set, so
existing deployments render identically.

No. The feature is opt-in and defaults preserve prior behavior.

- Unset vars → `helm template` renders no `OPENEBS_IO_CLIENT_QPS` /
  `OPENEBS_IO_CLIENT_BURST` env entries; client uses client-go defaults.
- Vars set via `localpv.controller.clientQPS` / `clientBurst` → both env vars
  rendered into the pod spec (verified for both Deployment and DaemonSet/node
  modes), and the provisioner logs the overridden QPS/burst at startup.
- Invalid / non-positive values → ignored, falls back to client-go defaults.

**Any additional information for your reviewer?**:

Continuation of #321 — that PR made the worker pool configurable; this PR makes
the shared client's rate limiter configurable so the added workers aren't
bottlenecked by the default client-side throttle. QPS is a `float32` and burst
an `int` to match the `rest.Config.QPS` / `rest.Config.Burst` field types in
client-go.

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Has the change log section been updated?
- [x] Commit has unit tests


cmd/provisioner-localpv/app/env.go
```diff
@@ -164,3 +180,35 @@ func getHelperPodTimeout() int {
...
+	val, present := menv.Lookup(ProvisionerClientQPS)
+	if !present || val == "" {
+		return 0, false
+	}
+	qps, err := strconv.ParseFloat(val, 32)
+	if err != nil || qps <= 0 {
+		return 0, false
+	}
+	return float32(qps), true
+}
```
niladrih, line 206: nit:
```suggestion
func getClientQPS() (float32, bool) {
	val, present := k8sEnv.GetFloat64(ProvisionerClientQPS, 0.0)
	if !present || val <= 0.0 {
		return 0.0, false
	}

	return float32(val), true
}
```

cmd/provisioner-localpv/app/env.go
```diff
@@ -164,3 +180,35 @@ func getHelperPodTimeout() int {
...
+func getClientBurst() (int, bool) {
+	val, present := menv.Lookup(ProvisionerClientBurst)
+	if !present || val == "" {
+		return 0, false
+	}
+	burst, err := strconv.Atoi(val)
+	if err != nil || burst <= 0 {
+		return 0, false
+	}
+	return burst, true
```
niladrih: nit:
```suggestion
    val, present := k8sEnv.GetInt(ProvisionerClientBurst, 0)
	if !present || val <= 0 {
		return 0, false
	}

	return val, true
```

cmd/provisioner-localpv/app/start.go
```diff
@@ -104,7 +104,20 @@ func Start(ctx context.Context, nodeDeployment bool, allowInsecurePvcBasePathOve
 	// OutOfCluster by using the following ENV variables:
 	//   OPENEBS_IO_K8S_MASTER - Kubernetes master IP address
 	//   OPENEBS_IO_KUBE_CONFIG - Path to the kubeConfig file.
-	kubeClient, err := mKube.New().Clientset()
+	// Apply client-side API rate-limit overrides only when explicitly set,
+	// so an unconfigured provisioner keeps the client-go defaults (QPS 5,
+	// Burst 10).
+	var clientOpts []mKube.OptionFn
```
niladrih: nit:
```suggestion
	clientOpts := make([]mKube.OptionFn, 0, 2)
```

pkg/kubernetes/client/client.go
```diff
@@ -235,6 +245,28 @@ func WithKubeConfigPath(kubeConfigPath string) OptionFn {
...
+}
+
+// WithBurst sets the maximum burst of queries above the QPS limit the client
+// may send to the Kubernetes API server. A non-positive value is ignored,
+// leaving the client-go default in place.
+func WithBurst(burst int) OptionFn {
+	return func(c *Client) {
+		if burst > 0 {
+			c.burst = burst
+		}
```
niladrih: Not required, as we're already doing this in applyRateLimits
```suggestion
		c.qps = qps
	}
}

// WithBurst sets the maximum burst of queries above the QPS limit the client
// may send to the Kubernetes API server. A non-positive value is ignored,
// leaving the client-go default in place.
func WithBurst(burst int) OptionFn {
	return func(c *Client) {
		c.burst = burst
```

deploy/helm/charts/values.yaml
```diff
@@ -46,6 +46,13 @@ localpv:
     # The number of concurrent worker goroutines for processing PVC create and delete events.
     # Higher values increase provisioning throughput when many PVC events are received simultaneously.
     workers: 4
+    # Maximum queries per second (QPS) the provisioner sends to the Kubernetes API server.
+    # Leave empty to use the client-go default (5). Raise it alongside `workers` to avoid
+    # client-side throttling when provisioning many volumes concurrently.
+    clientQPS: ""
+    # Maximum burst of queries above clientQPS the provisioner allows to the Kubernetes API
+    # server. Leave empty to use the client-go default (10).
+    clientBurst: ""
```
niladrih:
```suggestion
    client:
      qps: ""
      # Maximum burst of queries above clientQPS the provisioner allows to the Kubernetes API
      # server. Leave empty to use the client-go default (10).
      burst: ""
```

oneubauer:
Thank you for the feedback and revisions!

deploy/helm/charts/README.md
```diff
@@ -90,6 +90,8 @@ helm install openebs-localpv openebs-localpv/localpv-provisioner --namespace ope
 | `hostpathClass.reclaimPolicy`              | ReclaimPolicy for Hostpath PVs                                                                                                     | `"Delete"`                    |
 | `imagePullSecrets`                         | Provides image pull secrect                                                                                                        | `""`                          |
 | `localpv.controller.workers`               | The no. of concurrent worker goroutines for processing PVC create/delete events                                                    | `4`                           |
+| `localpv.controller.clientQPS`             | Max queries/sec to the Kubernetes API server. Empty uses the client-go default (5)                                                 | `""`                          |
+| `localpv.controller.clientBurst`           | Max burst of queries above clientQPS to the Kubernetes API server. Empty uses the client-go default (10)                           | `""`                          |
```
niladrih: This needs to reflect the changes to the values.yaml keys.

oneubauer: whoops; good catch. TY.

niladrih (approved):

niladrih:
@oneubauer could you please rebase your PR. Changes LGTM, will merge once you rebase.

niladrih:
@mergify squash

oneubauer:
> @oneubauer There are failing unit tests, could you please take a look?

Fixed. I took a deeper look and found other problems that I've corrected.

-  godocs said "a non-positive value is ignored," but the setters stored
whatever they were given — only `applyRateLimits` filtered. Added the guard to
the setters. This is what the failing unit tests were pointing at.

- (claude found this) A QPS-only override crashed the provisioner at startup.
`kubernetes.NewForConfig` rejects a config with `QPS > 0` and `Burst <= 0`
(client-go v0.34.1, `clientset.go:514`). Since `applyRateLimits` set each field
independently, `OPENEBS_IO_CLIENT_QPS=50` with burst unset produced exactly
that config, and `Start()` failed with:
```
    unable to get k8s client: burst is required to be greater than 0 when
    RateLimiter is not set and QPS is set to greater than 0
```    
So I fixed that too to help avoid a runtime fail.

niladrih:
@mergify rebase

niladrih:
@mergify squash

pkg/kubernetes/client/client.go
```diff
@@ -306,9 +340,35 @@ func (c *Client) Clientset() (*kubernetes.Clientset, error) {
...
+	if c.burst > 0 {
+		config.Burst = c.burst
+	}
+	// kubernetes.NewForConfig rejects a config with QPS > 0 and Burst <= 0,
+	// so a QPS-only override would fail client construction. Fall back to the
+	// burst client-go would otherwise have defaulted to, keeping the burst
+	// dimension unchanged from the caller's point of view.
+	if config.QPS > 0 && config.Burst <= 0 {
+		config.Burst = rest.DefaultBurst
+	}
```
niladrih, line 369: Will supplying Burst-only configuration work correctly?

niladrih (approved):

pkg/kubernetes/client/client.go
niladrih: I see there's a test for this.

niladrih:
@mergify queue

PR #352, opened 2026-07-09
test(client): fix inverted message and add missing assertion in TestConfig

anxkhn:

`TestConfig` in `pkg/kubernetes/client/client_test.go` (which exercises
`Client.Config()`) has a broken assertion block, so it both misreports on failure
and silently skips checks it is meant to perform:

1. The `if mock.isErr && err == nil` branch fires only when an error was expected
   but none occurred, yet its message reads `expected no error actual '%s'`, the
   inverse of the condition, and prints the nil `err` with `%s` (yielding
   `%!s(<nil>)`). Every sibling test in the file (`TestGetConfigFromENV`,
   `TestGetConfigFromPathOrDirect`, etc.) already uses the correct message
   `expected error actual no error` for the identical condition.

2. `TestConfig` is the only test in the file missing the success-direction check
   `if !mock.isErr && err != nil`. Because of that, the six table rows that
   declare `isErr: false` (`t1`, `t4`, `t5`, `t6`, `t7`, `t9`) never have their
   no-error expectation asserted. A regression that made `Config()` wrongly return
   an error on the in-cluster or ENV path would have kept the test green.

**What this PR does?**:

Replaces the single-branch assertion in `TestConfig` with the same two-branch form
already used by every sibling test in the file:

```go
if mock.isErr && err == nil {
    t.Fatalf("test '%s' failed: expected error actual no error", name)
}
if !mock.isErr && err != nil {
    t.Fatalf("test '%s' failed: expected no error actual '%s'", name, err)
}
```

Test-only change: 3 insertions, no production code touched.

No.

- `gofmt -l pkg/kubernetes/client/client_test.go` - clean.
- `go vet ./pkg/kubernetes/client/` - clean.
- `go test ./pkg/kubernetes/client/... -count=1` - PASS (all funcs, including
  `TestConfig` subtests `t1`-`t9`).
- Red -> green: temporarily pointing an `isErr: false` row (`t1`) at a stub that
  returns an error makes the new `!isErr && err != nil` branch fail the test;
  reverting turns it green again, confirming the added assertion catches a
  regression the previous test masked.

Pure test-hygiene fix, self-contained to `TestConfig`; it brings that test in line
with the assertion pattern the rest of the file already follows. No behavior of the
provisioner changes.

**Checklist:**
- [ ] Fixes # (no issue; test-only correctness fix)
- [ ] Has the change log section been updated? (N/A - test-only, not user-facing;
      matches precedent, e.g. docs PR #346 added no changelog entry)
- [x] Commit has unit tests (the change is the unit test correction itself)
- [ ] Commit has integration tests (N/A)
- [ ] (Optional) upgrade changes - N/A
- [ ] (Optional) docs changes - N/A


niladrih (approved):

niladrih (approved):
@anxkhn could you please rebase your branch. LGTM.

anxkhn:
@niladrih thanks for the approval. the branch is rebased onto the latest develop. is there anything else needed from me before this can be merged?

PR #306, opened 2025-12-10
Add Node-Deployment Mode for High-Performance LocalPV Provisioning

laik:
This PR introduces a node-deployment mode for the OpenEBS Dynamic LocalPV 
Provisioner, addressing performance bottlenecks identified in openebs/openebs#4050.

Key Changes:
- New `--node-deployment=true` flag to run provisioner as DaemonSet on each node
- Direct local volume operations (mkdir, quota, rm) without helper pods
- Automatic node filtering using IgnoredError for proper multi-instance coordination
- Leader election disabled in node-deployment mode

Benefits:
- Eliminates helper pod creation/termination overhead
- Reduces PVC binding latency significantly
- Decreases API server load
- Simpler architecture than external service approach

Solve xfs/ext projId Issues
Solve race conditions Issues

https://github.com/openebs/dynamic-localpv-provisioner/issues/289
https://github.com/openebs/dynamic-localpv-provisioner/pull/290

PR #355, opened 2026-07-14
fix(provisioner): honor FilePermissions mode in node-deployment mode

anxkhn:

In node-deployment (DaemonSet) mode the LocalPV hostpath provisioner ignores the
configured `FilePermissions` mode and always creates volume directories as
world-writable `0777`. The same StorageClass/PVC config is honored in the
default helper-pod mode, so the two modes disagree: a user who sets
`FilePermissions` to, say, `"0750"` or `"0700"` gets the requested mode with
helper pods but `0777` when the provisioner runs as a node DaemonSet. That is a
config regression with a permissions/security dimension (directories end up more
permissive than requested).

No tracking issue was filed; this is a mechanical regression found by reading the
node-deployment path. Happy to open one first if the maintainers prefer.

**What this PR does?**:

`ProvisionHostPath` already computes `fsMode` from `VolumeConfig.GetFsMode()`
(defaulting to `0777` when unset) and uses it to build the helper-pod
`mkdir -m <fsMode>` command. In node-deployment mode, though,
`createVolumeLocally` built its own `VolumeRequest` with `FsMode` hardcoded to
`"0777"` and never read that value, because `HelperPodOptions` did not carry the
mode.

This PR plumbs the already-computed `fsMode` through so both provisioning paths
share the same value:

- Add an `fsMode` field to `HelperPodOptions`.
- Set `fsMode: fsMode` in both `HelperPodOptions` constructions in
  `ProvisionHostPath` (the default and quota-enabled cases).
- Build the node-deployment `VolumeRequest` from `pOpts.fsMode` via a small
  `newLocalVolumeRequest` helper instead of the hardcoded `"0777"`.

`LocalVolumeManager.CreateVolume` already defaults an empty `FsMode` to `0777`,
so behavior is unchanged when `FilePermissions` is not set.

No. There are no CRD, API, config, or schema changes. When `FilePermissions` is
unset the behavior is identical to before (`0777`); the change only takes effect
for volumes that already opt into a specific mode.

Verified by unit test and static review of the request-construction path (I do
not have a live multi-node cluster with `--node-deployment` and the DaemonSet
host mount to exercise the full end-to-end flow):

- Added `TestNewLocalVolumeRequest` (table-driven): a configured mode (`0750`)
  and a restrictive mode (`0700`) are carried into the `VolumeRequest`, and an
  empty mode is left empty so the volume manager applies its `0777` default.
- Confirmed red -> green: with the previous hardcoded `"0777"` the new test fails
  all three cases; with the fix it passes.
- `gofmt -l`, `go vet`, `go build ./...`, and
  `go test ./cmd/provisioner-localpv/app/` are clean.

**Any additional information for your reviewer?**:

The `0777` default is preserved in two places on purpose (`ProvisionHostPath`
sets it when the config is empty, and `CreateVolume` re-defaults an empty
`FsMode`), so there is no behavior change for the unset case. The node-deployment
path was introduced in #306; this only fills the gap where its `VolumeRequest`
did not receive the configured mode. Not part of a larger design.

**Checklist:**
- [ ] Fixes #<issue number>  (no issue filed; mechanical regression, can open one if preferred)
- [ ] Has the change log section been updated?  (`changelogs/unreleased/` is
      currently empty and recent merged fix/feat PRs did not add entries; will add
      one if the maintainers want it)
- [x] Commit has unit tests
- [ ] Commit has integration tests  (needs a live node-DaemonSet cluster, not run here)
- [ ] (Optional) Are upgrade changes included in this PR? If not, mention the issue/PR to track:  n/a
- [ ] (Optional) If documentation changes are required, which issue on https://github.com/openebs/openebs-docs is used to track them:  n/a


niladrih (approved):

niladrih:
@mergify queue

